PR #13, opened 2017-09-25
Trip validation fix: compare trip stoptime stop IDs with pattern stop IDs

landonreed:
See #43. This fixes an issue where the pattern stops were not being compared against the trip stoptime stop IDs when creating a new trip.


landonreed:
agh... wrong repo

PR #15, opened 2017-11-21
Eliminate gtfs-api project

abyrd:
I cleaned up the gtfs-api project after our big JDBC / RDBMS refactor, removing all dead code. There were only a few classes left. I moved the ones that define the GraphQL schema and provide a String-based API into the `gtfs-lib` project and the HTTP layer here, into `datatools-server`. That way the HTTP layer is handled in the project that has the web framework as a dependency.


abyrd:
This PR was a mistake.

PR #17, opened 2018-12-07
Update repo URLs

landonreed:
Active development is returning to the Conveyal organization repo (see conveyal/datatools-ui#276). As such, we need to update the repo URL in order for automated releases to work correctly.


src/main/java/com/conveyal/gtfs/GraphQLMain.java
```diff
@@ -19,7 +19,7 @@
  * What we're trying to provide is this:
  * The queries that analysis-ui makes are at https://github.com/conveyal/analysis-ui/blob/dev/lib/graphql/query.js ;
  * note that feeds are wrapped in bundles in analysis-ui (we wrap the GTFS API types)
- * GraphQL queries for datatools-ui are at https://github.com/catalogueglobal/datatools-ui/blob/dev/lib/gtfs/util/graphql.js.
+ * GraphQL queries for datatools-ui are at https://github.com/conveyal/datatools-ui/blob/dev/lib/gtfs/util/graphql.js.
```
evansiroky: The queries are in more places than just here, but I don't think that oughta make this comment blocking the PR.

PR #64, opened 2018-12-10
Merge dev into master (release)

landonreed:
This PR will create the first release in the Conveyal repository in some time. The most recent datatools-server tagged release in the catalogueglobal repository was https://github.com/catalogueglobal/datatools-server/releases/tag/v3.2.0.

Before we merge to dev, let's add maven-semantic-release to the travis script, so we don't have to handle tagging the release manually (edit: this is done in #157).


landonreed:
Now that maven-semantic-release has been added, I'm going to merge to verify that the release system is working and because it has been so long since a release. Now that this is in place, I expect this project to have much more frequent releases.

Issue #152, opened 2018-12-17
MTC feed version publication cannot rely on md5 checksum


landonreed:
Recently, the MTC-specific code that:

1. copies a GTFS feed for a feed version to an external resource (S3 bucket),
2. waits for processing by an external application, and
3. looks for the processed feed in a new folder in that S3 bucket

was refactored to use md5 checksums to determine when the GTFS file for a specific version had completed processing. The assumption that the file was unchanged throughout this process was invalid, so we must use some other means to check this. There doesn't seem to be any reliable method to trace a file back to a specific version, so for now we'll have to just assume that the latest user-published file must be the file that ends up in the "completed processing" folder.

PR #153, opened 2018-12-17
Fixes for the MTC migration to RDBMS

landonreed:
This PR covers a few items:

- This issue references #152 by replacing the MD5 comparison with a check for the latest published version for a feed source, but doesn't entirely solve the problem so I will leave it open for the time being.
- Fixes an issue with pattern stop migration from MapDB where the sequence number was not zero-based.
- Enhances `UpdateSQLFeedsMain` by
   - permitting a list of values to be passed in for applying a SQL statement patch to sets of feeds in the GTFS database.
  - clarifying `null` value as arg
  - adds option to run SQL statement as test run (with rollback)
- Fixes a minor bug where deployment to s3 is disabled if s3 config set to false.


src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -192,10 +192,10 @@ public FeedValidationResultSummary validationSummary() {
     public String namespace;
 
     /**
-     * Indicates whether a feed version is pending published status, a check that is currently performed in
-     * {@link com.conveyal.datatools.manager.jobs.FeedUpdater}. This field is currently in use only for the MTC extension.
+     * Indicates when (if at all) a feed version was published to an external source. This field is currently in use
+     * only for the MTC extension and is reset to null after the published version has been registered externally.
      * */
-    public boolean processing;
+    public Date published;
```
evansiroky: Can we rename this variable to `externallyPublished`?  Also, I don't quite understand why it's being set to null.

landonreed: @evansiroky, yes we can change that variable name. As for why it's being set to null, there are three states a version can be for the UI: published/processed by the external source, which is indicated by `FeedSource#publishedVersionId`; processing, which is indicated by a non-null value in this field (`FeedVersion#externallyPublished`); and not published, which is indicated by a null value in this field.

I figured using null values here was better than leaving the date value intact and requiring the UI to compare these values to find the most recent (which would indicate which version is currently processing). I do realize though that this is a bit of a strange pattern. Do you have any thoughts on improving it?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedVersionController.java
```diff
@@ -272,17 +270,23 @@ private static FeedVersion publishToExternalResource (Request req, Response res)
...
+                Persistence.feedSources.updateField(version.feedSourceId, "publishedVersionId", version.namespace);
+                return version;
+            } else {
+                // NOTE: If the MTC extension is enabled, the parent feed source's publishedVersionId will not be updated to the
+                // version's namespace until the FeedUpdater has successfully downloaded the feed from the share S3 bucket.
+                Date publishedDate = new Date();
+                return Persistence.feedVersions.updateField(version.id, "published", publishedDate);
+            }
+        } catch (Exception e) {
+            haltWithMessage(req, 400, "Could not publish feed.", e);
```
evansiroky: This haltWithMessage should probably return a 500 error.  If the exception is actually due to bad user input, we should try to return a more helpful error message.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -91,36 +95,31 @@ public void run() {
...
-                    LOG.warn("New version found for at {}/{}. ETag = {}. Downloading from s3", feedBucket, keyName, eTag);
-                    S3Object object = FeedStore.s3Client.getObject(feedBucket, keyName);
-                    InputStream in = object.getObjectContent();
-                    File file = new File(FeedStore.basePath, filename);
-                    OutputStream out = new FileOutputStream(file);
-                    ByteStreams.copy(in, out);
-                    String md5 = HashUtils.hashFile(file);
+                    LOG.info("New version found for {} at s3://{}/{}. ETag = {}.", feedId, feedBucket, keyName, eTag);
                     FeedSource feedSource = null;
                     List<ExternalFeedSourceProperty> properties = Persistence.externalFeedSourceProperties.getFiltered(and(eq("value", feedId), eq("name", AGENCY_ID)));
```
evansiroky: Please make this expand to a few lines to be more readable.  Also, see https://github.com/conveyal/datatools-server/pull/153/files#r245153439.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -91,36 +95,31 @@ public void run() {
...
-        // iterate over feeds in download_prefix folder and register to gtfsApi (MTC project)
+        // iterate over feeds in download_prefix folder and register to (MTC project)
         ObjectListing gtfsList = FeedStore.s3Client.listObjects(feedBucket, bucketFolder);
+        LOG.info(eTagForFeed.toString());
         for (S3ObjectSummary objSummary : gtfsList.getObjectSummaries()) {
 
             String eTag = objSummary.getETag();
+            String keyName = objSummary.getKey();
+            LOG.info("{} etag = {}", keyName, eTag);
             if (!eTagForFeed.containsValue(eTag)) {
```
evansiroky: This `eTagForFeed` map variable never uses the key except possibly for logging/debugging purposes.  Consider making this variable a Set instead.

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
evansiroky: One idea for improvement would be to change the variable name to something like `timeSentToExternalPublisher` then create another variable of something like `timeWhenExternalPublisherAppearedToHaveFinishedProcessingFeedVersion`.  I don't think overwriting the variable to null is a good idea.  First, it erases information.  Second, when the FeedVersion `published` variable is null, then that has meaning attached to it.  I think it would be a lot better to set another variable to a non-null state to convey a positive state.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -6,13 +6,17 @@
...
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
```
evansiroky: Not a comment about this line, but the class javadoc needs to be updated.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
landonreed: I have found it helpful to have this as a map during debugging/logging, so I think we should keep it that way.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -166,21 +173,17 @@ private void updatePublishedFeedVersion(String feedId, FeedSource feedSource) {
             // feed that failed processing by RTD.
             Optional<FeedVersion> lastPublishedVersionCandidate = versions
                 .stream()
-                .min(Comparator.comparing(v -> v.published, Comparator.nullsLast(Comparator.reverseOrder())));
+                .min(Comparator.comparing(v -> v.sentToExternalPublisher, Comparator.nullsLast(Comparator.reverseOrder())));
             if (lastPublishedVersionCandidate.isPresent()) {
                 FeedVersion publishedVersion = lastPublishedVersionCandidate.get();
-                if (publishedVersion.published == null) {
+                if (publishedVersion.sentToExternalPublisher == null) {
                     LOG.warn("Not updating published version for {} (published date is null)", feedId);
```
evansiroky: This log statement could be updated to note not being sent to the external publisher.

PR #157, opened 2019-01-04
Add maven-semantic-release

landonreed:
Add https://github.com/conveyal/maven-semantic-release so that each merge to master will produce a new release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
 # package jar
 - mvn package
 after_success:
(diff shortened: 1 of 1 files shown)


evansiroky (changes requested):
This is not enough to get it to work since maven-semantic-release expects more things to be setup to release to maven-central.  We'd need to add a `maven-settings.xml` file and also all the gpg keys.  If we don't want to release to maven-central, maven-semantic-release might actually not be the right tool for this job.

evansiroky:
[maven-semantic-release v4.2.0](https://www.npmjs.com/package/@conveyal/maven-semantic-release/v/4.2.0) now allows releasing maven projects without deploying to maven central.

evansiroky (approved):

PR #168, opened 2019-01-28
Fix gtfs plus upload

landonreed:
Fixes GTFS+ upload and publication (version creation). Basically, the GTFS+ code was never fully updated for the Mongo + RDBMS changes, so this seeks to finish that work. This is why, for example, it may appear strange that the ProcessSingleFeedJob is being added to the publish controller, when it didn't previously exist.

This works in conjunction with conveyal/datatools-ui#382.

It also adds stop count to validation summary.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/GtfsPlusController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/GtfsPlusController.java
@@ -13,20 +13,12 @@ import com.conveyal.datatools.manager.utils.json.JsonUtil;
 import com.conveyal.gtfs.GTFSFeed;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.google.common.io.ByteStreams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
 
-import javax.servlet.MultipartConfigElement;
-import javax.servlet.ServletException;
-import javax.servlet.ServletInputStream;
-import javax.servlet.ServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
-import javax.servlet.http.Part;
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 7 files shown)


src/main/java/com/conveyal/datatools/manager/controllers/api/GtfsPlusController.java
```diff
@@ -162,7 +141,7 @@ private static Long getGtfsPlusFileTimestamp(Request req, Response res) {
         }
     }
 
-    private static Boolean publishGtfsPlusFile(Request req, Response res) {
+    private static String publishGtfsPlusFile(Request req, Response res) {
```
evansiroky: Please add comments to what this method is supposed to do.

src/main/java/com/conveyal/datatools/manager/controllers/api/GtfsPlusController.java
```diff
@@ -55,30 +54,10 @@
     private static FeedStore gtfsPlusStore = new FeedStore("gtfsplus");
 
 
-    public static Boolean uploadGtfsPlusFile (Request req, Response res) throws IOException, ServletException {
-
-        //FeedSource s = FeedSource.retrieveById(req.queryParams("feedSourceId"));
+    private static Boolean uploadGtfsPlusFile (Request req, Response res) {
```
evansiroky: Please add some comments to this method explaining what it does.  Also, please add some comments to the class in addition to the original author.

src/main/java/com/conveyal/datatools/manager/controllers/api/GtfsPlusController.java
```diff
@@ -230,21 +208,25 @@ private static Boolean publishGtfsPlusFile(Request req, Response res) {
...
 
-        newFeedVersion.hash();
-
-        // validation for the main GTFS content hasn't changed
-        newFeedVersion.validationResult = feedVersion.validationResult;
-        newFeedVersion.storeUser(profile);
-        Persistence.feedVersions.create(newFeedVersion);
+        // Must be handled by executor because it takes a long time.
+        ProcessSingleFeedJob processSingleFeedJob = new ProcessSingleFeedJob(newFeedVersion, profile.getUser_id(), true);
+        DataManager.heavyExecutor.execute(processSingleFeedJob);
```
evansiroky: The whole workflow of this method is unclear to me.  Why is the job being added when it wasn't there before?  This whole class could use a few more comments and this PR could use an issue explaining what is being fixed here.

evansiroky:
This PR is now waiting on @abyrd to comment on https://github.com/conveyal/datatools-server/pull/169#issuecomment-460437453.  After that happens and if we decide to merge #170 into this, I'll give approval.

landonreed:
Pending changes from @evansiroky in #170.

Issue #166, opened 2019-01-07
Scheduler#scheduleExpirationNotifications concurrency issue


landonreed:
During a import of a large number of feed, I ran into this exception:

```
Jan 07 19:42:46.196 [pool-1-thread-4] ERROR c.c.d.common.status.MonitorableJob:180 - Job failed
java.util.ConcurrentModificationException: null
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901) ~[na:1.8.0_91]
	at java.util.ArrayList$Itr.next(ArrayList.java:851) ~[na:1.8.0_91]
	at com.google.common.collect.AbstractMapBasedMultimap$WrappedCollection$WrappedIterator.next(AbstractMapBasedMultimap.java:486) ~[dt-latest-dev.jar:na]
	at com.conveyal.datatools.common.utils.Scheduler.removeJobsOfType(Scheduler.java:86) ~[dt-latest-dev.jar:na]
	at com.conveyal.datatools.common.utils.Scheduler.removeFeedSourceJobsOfType(Scheduler.java:100) ~[dt-latest-dev.jar:na]
	at com.conveyal.datatools.common.utils.Scheduler.scheduleExpirationNotifications(Scheduler.java:181) ~[dt-latest-dev.jar:na]
	at com.conveyal.datatools.manager.jobs.ValidateFeedJob.jobFinished(ValidateFeedJob.java:52) ~[dt-latest-dev.jar:na]
	at com.conveyal.datatools.common.status.MonitorableJob.run(MonitorableJob.java:170) [dt-latest-dev.jar:na]
	at com.conveyal.datatools.common.status.MonitorableJob.run(MonitorableJob.java:145) [dt-latest-dev.jar:na]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_91]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [na:1.8.0_91]
	at java.lang.Thread.run(Thread.java:745) [na:1.8.0_91]
Jan 07 19:42:46.196 [pool-1-thread-4] INFO  c.c.d.common.status.MonitorableJob:185 - VALIDATE_FEED 2e90f69e-1b43-44c2-a6df-f5408a4ca6ab errored in 1501 ms
[...]

PR #170, opened 2019-02-02
Upload job fixes

evansiroky:
I noticed some errors happening while testing out the GTFS Plus fixes.  This should fix an error with removing scheduled jobs from a feed and also an error that occurs on feeds that don't have a url, but are set to automatically fetch.


landonreed:
@abyrd, @evansiroky is referring to the errors described in #166.

evansiroky:
I think this should resolve the matter of making the ListMultiMaps thread-safe.

PR #176, opened 2019-02-12
User controller refactor

evansiroky:
### Checklist

### Description

The user controller has a lot of repetitive code and does not have thorough logging. This PR consolidates a lot of common code and adds many more logging statements to help with debugging. This should work with a config with valid auth0 settings.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
@@ -166,7 +166,7 @@ public class UserController {
         return executeRequestAndGetResult(request, req);
     }
 
-    private static Object updateUser(Request req, Response res) {
+    private static String updateUser(Request req, Response res) {
         String userId = req.params("id");
         Auth0UserProfile user = getUserById(userId);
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -327,6 +282,140 @@ private static Object getRecentActivity(Request req, Response res) {
...
+    /**
+     * Create a HTTP POST request with some default headers
+     */
+    private static HttpPost makePostRequest(String url) {
+        HttpPost request = new HttpPost(url);
+        request.addHeader("Authorization", "Bearer " + AUTH0_API_TOKEN);
+        request.setHeader("Accept-Charset", charset);
+        request.setHeader("Content-Type", "application/json");
+        return request;
+    }
```
landonreed: It seems like there is a lot of duplicated code in `makePostRequest` and `makeGetRequest`. Also, the same headers are set in `deleteUser`. Perhaps, this should all be refactored into a single function that takes as input an `HttpEntityEnclosingRequestBase` and sets headers the Auth0-related headers called, for example, `setAuth0Headers`. It probably wouldn't hurt to set content type for all request types just to simplify things (or you could only set that header for posts). Thoughts?

evansiroky:
Just consolidated code as suggested.

landonreed:
Thanks for making that change. I guess I'm unclear as to why `makeGetRequest` and `makePostRequest` would still need to exist now that we have `setHeaders`. At this point they just seem to be adding bloat to the code. I suppose technically `makePostRequest` is used twice, but it seems better to just have each controller method handle creating the HTTP request the same way.

evansiroky:
Turns out it saves lines of code to not have those helper functions now.

PR #184, opened 2019-02-15
Log response body when an error from auth0 is received

evansiroky:
### Checklist

- [ ] All tests and CI builds passing

### Description

While investigating an issue of a failing Auth0 operation, I needed to view the response body from Auth0 to discover what the error message was. This PR seeks to add logging of the response body when a failed Auth0 API call occurs.


landonreed:
@evansiroky, I'll note here for example's sake that this PR is decreasing the code coverage for the project. When it eventually gets merged to master, the checklist item requiring an increase in code coverage will not be met. This is why I don't think it's a good idea to strictly enforce an ever-increasing code coverage requirement for PRs.

PR #187, opened 2019-02-16
Auth0 error response improvements

evansiroky:
### Checklist

### Description

- When an Auth0 error is encountered while creating, updating or deleting a user, the message of the Auth0 error will be passed through properly. This should specifically help when a user tries to create a new user with the same email as an existing user.
- Adds more logging to various aspects of communicating with the Auth0 API.
- Adds tests for creating, reading, updating and deleting users
- Downgrades junit from v5 to v4 in order to make it compatible with Wiremock. The only reason we upgraded to junit 5 was due to some comments from @abyrd in https://github.com/catalogueglobal/datatools-server/pull/115#discussion_r218301236. However, [`BeforeAll` in v5 is basically the same thing as `BeforeClass` in v4](https://stackoverflow.com/a/20295618/269834). [Wiremock is not compatible with Junit 5](https://github.com/tomakehurst/wiremock/issues/684), but it provides such great functionality that I think the benefits of upgrading to junit 5 are heavily outweighed by the downside of not being able to use Wiremock.


evansiroky:
I have removed all tests except for the one test that demonstrates that the updated functionality is able to return a meaningful error given the auth0 response for a duplicate user.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Connection.java
```diff
@@ -132,8 +135,8 @@ private static void remapTokenValues(Map<String, Object> jwt) {
      * tables in the database.
      */
     public static void checkEditPrivileges(Request request) {
-        if (authDisabled()) {
-            // If in a development environment, skip privileges check.
+        if (authDisabled() || IN_TESTING_ENVIRONMENT) {
+            // If in a development or testing environment, skip privileges check.
```
landonreed: Why do we want to skip the edit privileges check here? Wouldn't it be valuable to check that the test user has editing privileges for a given feed source during the course of testing. Maybe there is a good reason, but it isn't obvious from the comment.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -59,9 +59,13 @@
...
     private static final String charset = "UTF-8";
-    private static String baseUsersUrl = "https://" + AUTH0_DOMAIN + "/api/v2/users";
-    public static JsonManager<Project> json =
-            new JsonManager<>(Project.class, JsonViews.UserInterface.class);
+    // set the base users URL based on whether the domain is the test environment
+    private static String baseUsersUrl = (
+        AUTH0_DOMAIN.equals("localhost:8089")
+            ? "http://localhost:8089"
+            : "https://" + AUTH0_DOMAIN
+    ) + "/api/v2/users";
```
landonreed: I feel like the test code is leaking into the main code a little bit here. Is there a better way to isolate the test URL or override the default value from the test code?

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -174,6 +176,10 @@ private static String updateUser(Request req, Response res) {
         String userId = req.params("id");
         Auth0UserProfile user = getUserById(userId);
 
+        if (user == null) {
+            logMessageAndHalt(req, 404, "User not found");
```
landonreed: I feel like this message might be a little vague given the UserController context. Maybe it should clarify which user was not found (the user to be created vs. the creating/requesting user).

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        // make request and parse the json response
+        JsonNode createUserResponse = parseJson(
+            given()
+                .port(4000)
+                .body("{\n" +
+                    "  \"email\" : \"test-existing-user@test.com\",\n" +
+                    "  \"password\" : \"password\",\n" +
+                    "  \"permissions\" : {}\n" +
+                    "}")
```
landonreed: I recognize that we use the same JSON string construction in the UserController, but I think we ought to move in the direction of constructing strings using JsonNodes both in this test class and in the controller. For example:

```
ObjectNode requestJson = mapper.createObjectNode();
requestJson.put("email", "test-existing-user@test.com");
requestJson.put("password", "password");
requestJson.putObject("permissions");
```

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Make sure a meaningful Auth0 error can be returned when a duplicate user is being created
+     */
+    @Test
+    public void canReturnMeaningfulAuth0Error() throws IOException {
+        // create wiremock stub for create users endpoint that responds with a message saying a user with the email
+        // already exists
+        stubFor(
+            post(urlPathEqualTo("/api/v2/users"))
+                .withRequestBody(matchingJsonPath("$.email", equalTo("test-existing-user@test.com")))
```
landonreed: Since this email string is used below, it seems like a good idea to place this into a constant and reuse.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+ * certain responses needed to verify functionality.
+ */
+public class UserControllerTest {
+    /**
+     * This sets up a mock server that accepts requests and sends predefined responses to mock an Auth0 server.
+     */
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(
+        options()
+            .port(8089)
```
landonreed: I don't love that the port is set separately from the `AUTH0_DOMAIN` value in the test env.yml. I also don't love the proliferation of config files, especially when only two values are changed in order to test the user controller. Maybe an alternative to simplify things would be to add a `configureAuth0` method in the UserController. We could call this within the `register` method with the values from the config, but then recall it in the `UserControllerTest#setUp` method (after `DataToolsTest.setUp()` of course) to override the values with `testing-client-id` and `localhost:8089`. That way the testing stuff doesn't bleed into the controller as mentioned in my previous comment.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        // make request and parse the json response
+        JsonNode createUserResponse = parseJson(
+            given()
+                .port(4000)
+                .body("{\n" +
+                    "  \"email\" : \"test-existing-user@test.com\",\n" +
+                    "  \"password\" : \"password\",\n" +
+                    "  \"permissions\" : {}\n" +
+                    "}")
+                .post("/api/manager/secure/user")
```
landonreed: It seems like a good idea to reuse the `DataManager.API_PREFIX` constant here (and in other tests).

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
abyrd: I definitely like the idea of constructing data with typed objects instead of embedding strings.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Connection.java
evansiroky: Hmmm, that's not a bad point. Since this was the first draft of testing an endpoint that required authentication, I just disabled it checking all privileges entirely so that any endpoint requiring authentication could be tested. I suppose the more thorough way of doing this would be to send an actual token that verifies with an actual AUTH0_SECRET or AUTH0_PUBLIC_KEY config value. However, that could conceivably be accomplished by writing a unit test for this class that had bad keys and bad tokens and whatnot and found the proper errors. 

Perhaps the solution here would be to create an actual Auth0 tenant, get the keys for testing and then delete the tenant so that those keys then aren't able to do anything. Lmk if that sounds like a good plan.

landonreed: For now, I would just say that we should bulk up the comment with a little more detail on why we're skipping the check (maybe along with a `TODO`). I don't think we need to go down the path of creating and deleting an Auth0 tenant. My long-term, preferred approach might be to assign a more complex `Auth0UserProfile` to the `user` request attribute that has certain permissions that we want to test.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
evansiroky: Just pushed some changes that hopefully address this comment.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
evansiroky: Added more context to this message in a recent commit.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
evansiroky: Thanks for the example!

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
evansiroky: Done

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
evansiroky: Just pushed some changes that hopefully address this comment.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
evansiroky: I pushed a change that does this, but it doesn't look as pretty anymore. See https://github.com/conveyal/datatools-server/pull/187/files#diff-f944fd66f929625958fa37e80336a2ebR96

evansiroky:
Addressed most of these comments, lmk what you think, @landonreed.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Connection.java
```diff
@@ -132,8 +134,11 @@ private static void remapTokenValues(Map<String, Object> jwt) {
      * tables in the database.
      */
     public static void checkEditPrivileges(Request request) {
-        if (authDisabled()) {
-            // If in a development environment, skip privileges check.
+        if (authDisabled() || inTestingEnvironment) {
+            // If in a development or testing environment, skip privileges check. This is done so that basically any API
+            // endpoint can function.
+            // TODO: make unit tests of the below items or do some more stuff as mentioned in PR review here:
+            // https://github.com/conveyal/datatools-server/pull/187#discussion_r262714708
```
landonreed: I think it's often good to link to issues in the code, but in this case I think we should just state what we want to do directly in the `TODO`: `TODO: make unit tests of the below permissions checks. This would require creating one or more complete mock users (Auth0UserProfile) that contains a full set of permissions.`

src/main/java/com/conveyal/datatools/manager/auth/Auth0Connection.java
```diff
@@ -188,4 +193,12 @@ public static void checkGTFSPrivileges(Request request) {
             }
         }
     }
+
+    /**
+     * This is used to activate behavior within this class that emulates a testing environment. Obviously, this is used
+     * and should only be used while running tests of this project.
+     */
+    public static void setInTestingEnvironment (boolean inTestingEnvironment) {
```
landonreed: Instead of using this method and the inTestingEnvironment  boolean, I think it might be cleaner to stash the testing url (localhost:8089) in a public constant in the test class and use a `isTestEnvironment` method that checks if the base URL equals the test URL. Really any changes we can make to reduce the number of times string literals are redefined or having to call multiple functions to do the same thing.

evansiroky: I'm don't understand what you've got in mind here. It seems like this would mean bringing back [this code](https://github.com/conveyal/datatools-server/pull/187#discussion_r261688311) you wanted removed.

landonreed: Sorry for the confusion, I'll create a PR.

PR #220, opened 2019-08-01
Properly detect rows with incorrect number of fields

landonreed:
### Checklist

### Description

Changes the GTFS+ validation so that it properly detects when files have rows that do not have the correct number of fields. E.g. 
[dumbarton_express_wrong_num_fields.zip](https://github.com/ibi-group/datatools-server/files/3456955/dumbarton_express_wrong_num_fields.zip)

This also fixes an issue where an input file had a BOM that was not being properly ignored: 
[SolTrans-20190716T195942Z-d2d170bb-18bc-44f7-8569-96d8a1de309b.zip](https://github.com/ibi-group/datatools-server/files/3456970/SolTrans-20190716T195942Z-d2d170bb-18bc-44f7-8569-96d8a1de309b.zip)

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/gtfsplus/GtfsPlusValidation.java
+++ b/src/main/java/com/conveyal/datatools/manager/gtfsplus/GtfsPlusValidation.java
@@ -8,6 +8,7 @@ import com.conveyal.datatools.manager.persistence.Persistence;
 import com.conveyal.gtfs.GTFSFeed;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import org.apache.commons.io.input.BOMInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/manager/gtfsplus/GtfsPlusValidation.java
```diff
@@ -133,21 +137,34 @@ private static void validateTable(
...
+                // If value exists for index, use that. Otherwise, default to null to avoid out of bounds exception.
+                String val = f < values.length ? values[f] : null;
+                validateTableValue(issues, tableId, rowIndex, val, fieldsFound[f], gtfsFeed);
             }
             rowIndex++;
         }
+        // Add issue for wrong number of columns after processing all rows.
+        if (rowsWithWrongNumberOfColumns > 0) {
+            issues.add(new ValidationIssue(tableId, null, -1, rowsWithWrongNumberOfColumns + " row(s) do not contain the same number of fields as there are headers. (File may need to be edited manually.)"));
+        }
```
evansiroky: I think it'd be better to create an issue for each row that has the incorrect amount of fields instead of just one per table.

landonreed: Yes, I considered this, but the reason I opted for this approach was that there's no concept of a row-level issue in the UI right now. So we would potentially need to add that to the UI somewhere. Also, there's the trouble of reporting the issue at the row level, but not really giving the user a great way to resolve the issue in the GTFS+ editor. Essentially, all of the rows with the wrong number of columns can be resolved simply by clicking the "Save and Revalidate" button -- so the resolution is more at the table level than the row level (like, for example, a bad value for a field would be). Not sure if all of that makes sense...

evansiroky: Hmm, ok, sounds reasonable.

landonreed: I'll copy some version of this conversation into the code to avoid confusing someone later on.

PR #225, opened 2019-08-08
Deploy OTP to AWS load balancer and manage OTP servers in separate collection (and misc other fixes)

landonreed:
### Checklist

### Description

Note: this contains code originally contained in catalogueglobal#124

We have come to realize that the way we deploy new graphs to OTP (making a "build over the wire" request to existing EC2 servers) isn't reliable at all. This PR in conjunction with a UI PR adds to Data Tools the option to set up a "server" that targets a load balancer (rather than an existing EC2 server). It will:

- spin up a new EC2 instance,
- have the user script build the graph,
- optionally upload the graph to S3, and finally
- run OTP.

Another important change here that is related to deployments is to move the OtpServer objects stored in the Project POJO to their own MongoDB collection. Now, these objects are tied to projects by projectId and fetched via a JsonProperty-annotated getter method. This additionally allows for one of these server targets to be tied to a project or (if projectId is null) available across all projects in the application.

Finally, this PR does some other housekeeping:

fixes catalogueglobal#120
removes unused/legacy snapshot classes
fixes catalogueglobal#110
fixes catalogueglobal/datatools-ui#234

### Notes on configuration
Instructions for setting up EC2 auto deploy:

    1. Create a new application load balancer in the desired VPC.
    2. Create a target group for the load balancer (arn is defined in OTP server)
    3. Ensure you have an AMI with the appropriate software installed and nginx configuration (set at `deployment.ec2.default_ami`).
    4. Create an OTP Server at http://localhost:9966/admin/servers with `Use Elastic Load Balancer (ELB)?` checked. Make sure to add required fields (Subnet ID, Security Group ID, IAM Role ARN, Key file name (without .pem), Target Group ARN) to the server.
    6. Deploy a feed to your new server!

Deployment module in `server.yml` should look like:

```
  deployment:
    enabled: true
    ec2:
      enabled: true
      default_ami: ami-041ee0ca5cd75f7d7
```


src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -148,10 +160,41 @@ private static void validateFields(Request req, Document serverDocument) throws
...
+            if (!serverDocument.containsKey("internalUrl") || ((List) serverDocument.get("internalUrl")).size() == 0) {
+                logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Server must contain either internal URL(s) or s3 bucket name.");
+            }
+        } else {
+            // Verify that application has permission to write to/delete from S3 bucket.
+            String key = UUID.randomUUID().toString();
+            String bucket = serverDocument.get("s3Bucket").toString();
+            try {
+                FeedStore.s3Client.putObject(bucket, key, File.createTempFile("test", ".zip"));
+                FeedStore.s3Client.deleteObject(bucket, key);
```
evansiroky: Is there an alternative way to check this? This will incur a monetary cost (probably unnoticeable, but more than $0) to put and delete an object.

landonreed: This was the recommended way I found: https://stackoverflow.com/a/17284647/915811

There might be an alternative, but this certainly seems like the most reliable way to handle this. I worry about introducing any complexity to optimize this, but if there's some other simple way, I'm definitely open to it.

evansiroky: Ok, cool, let's keep it as is, but please put that stackoverflow link in the code comments.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,213 @@
...
+            ? userProfile.canAdministerApplication()
+            : userProfile.canAdministerProject(projectId, organizationId);
+        if (allowedToCreate) {
+            validateFields(req, newServerFields);
+            try {
+                OtpServer newServer = mapper.readValue(newServerFields.toJson(), OtpServer.class);
+                Persistence.servers.create(newServer);
+                return newServer;
+            } catch (IOException e) {
+                logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Error parsing OTP server JSON.");
```
evansiroky: Please add the `e` variable to the `logMessageAndHalt` call.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,213 @@
...
+            try {
+                DescribeTargetGroupsRequest describeTargetGroupsRequest =
+                    new DescribeTargetGroupsRequest().withTargetGroupArns(serverDocument.get("targetGroupArn").toString());
+                AmazonElasticLoadBalancing elb = AmazonElasticLoadBalancingClient.builder().build();
+                List<TargetGroup> targetGroups = elb.describeTargetGroups(describeTargetGroupsRequest).getTargetGroups();
+                if (targetGroups.size() == 0) {
+                    logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Invalid value for Target Group ARN. Could not locate Target Group.");
+                }
+            } catch (AmazonElasticLoadBalancingException e) {
+                logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Invalid value for Target Group ARN.");
```
evansiroky: Please add the `e` variable to the `logMessageAndHalt` call.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,213 @@
...
+            }
+            // Validate instance type
+            if (serverDocument.get("instanceType") != null) {
+                try {
+                    InstanceType.fromValue(serverDocument.get("instanceType").toString());
+                } catch (IllegalArgumentException e) {
+                    String message = String.format("Must provide valid instance type (if none provided, defaults to %s).", DEFAULT_INSTANCE_TYPE);
+                    logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, message);
```
evansiroky: Please add the `e` variable to the `logMessageAndHalt` call.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,213 @@
...
+            // miniscule).
+            String key = UUID.randomUUID().toString();
+            String bucket = serverDocument.get("s3Bucket").toString();
+            try {
+                FeedStore.s3Client.putObject(bucket, key, File.createTempFile("test", ".zip"));
+                FeedStore.s3Client.deleteObject(bucket, key);
+            } catch (IOException | AmazonS3Exception e) {
+                String message = "Cannot write to specified S3 bucket " + bucket;
+                LOG.error(message, e);
+                logMessageAndHalt(req, 400, message);
```
evansiroky: Please add the `e` variable to the `logMessageAndHalt` call.

src/main/java/com/conveyal/datatools/manager/controllers/api/ProjectController.java
```diff
@@ -120,10 +120,7 @@ private static Project updateProject(Request req, Response res) {
     private static Project deleteProject(Request req, Response res) {
         // Fetch project first to check permissions, and so we can return the deleted project after deletion.
         Project project = requestProjectById(req, "manage");
-        boolean successfullyDeleted = project.delete();
-        if (!successfullyDeleted) {
-            logMessageAndHalt(req, 500, "Did not delete project.", new Exception("Delete unsuccessful"));
-        }
+        project.delete();
```
evansiroky: Just curious why this was changed.

landonreed: There's not a great reason. If I recall correctly, there were some merge conflicts related to the `Project#delete` method with the old `deploy-to-ec2` branch. I think my thinking was that this wouldn't be providing very valuable information about what went wrong (it was just returning a boolean from the mongo delete operation).

configurations/default/server.yml.tmp
```diff
@@ -13,6 +13,15 @@ modules:
     enabled: false
   editor:
     enabled: false
+  deployment:
+    enabled: false
+    ec2:
+      enabled: false
+      subnet: subnet-id
+      securityGroup: security-group-id
```
evansiroky: Are we sure that this is information we'd want to include as part of the `appinfo` response?

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -114,56 +164,79 @@ public void jobLogic () {
...
+            // If we come to this point, there are internal URLs we need to deploy to (i.e., build graph over the wire).
+            boolean sendOverWireSuccessful = buildGraphOverWire();
+            if (!sendOverWireSuccessful) return;
+            // Set baseUrl after success.
+            status.baseUrl = otpServer.publicUrl;
         }
+        status.completed = true;
+    }
+
+    private void uploadBundleToS3() throws InterruptedException, AmazonClientException {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
         NotifyUsersForSubscriptionJob.createNotification("deployment-updated", deployment.id, message);
     }
 
+    public void replaceEC2Servers() {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -114,56 +164,79 @@ public void jobLogic () {
...
+        String copyKey = getLatestS3BundleKey();
+        CopyObjectRequest copyObjRequest = new CopyObjectRequest(s3Bucket, key, s3Bucket, copyKey);
+        FeedStore.s3Client.copyObject(copyObjRequest);
+        LOG.info("Copied to s3://{}/{}", s3Bucket, copyKey);
+        LOG.info("Uploaded to s3://{}/{}", s3Bucket, getS3BundleKey());
+        status.update("Upload to S3 complete.", status.percentComplete + 10);
+        status.uploadingS3 = false;
+    }
 
+    private boolean buildGraphOverWire() {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+                    MonitorServerStatusJob monitorServerStatusJob = new MonitorServerStatusJob(owner, deployment, instance, otpServer);
+                    service.submit(monitorServerStatusJob);
+                }
+                // Shutdown thread pool once the jobs are completed and wait for its termination. Once terminated, we can
+                // consider the servers up and running (or they have failed to initialize properly).
+                service.shutdown();
+                service.awaitTermination(4, TimeUnit.HOURS);
+            }
+            String finalMessage = "Server setup is complete!";
+            if (otpServer.instanceIds != null) {
```
evansiroky: This looks like the default behavior for deployment involves decommissioning other ec2 instances. This wasn't written in the tutorial and I think it's important to note in case someone is just trying things out and overwrites an existing deployment. The UI ought to have some kind of check or warning describing how many instances are currently running.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+            Persistence.servers.updateField(otpServer.id, "instanceIds", getIds(instances));
+            // Job is complete? FIXME Do we need a status check here?
+            status.update(false, finalMessage, 100, true);
+        } catch (Exception e) {
+            LOG.error("Could not deploy to EC2 server", e);
+            status.fail("Could not deploy to EC2 server", e);
+        }
+    }
+
+    private List<Instance> startEC2Instances(int count) {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+        }
+        LOG.info("Public IP addresses have all been assigned. {}", instanceIpAddresses.values().toString());
+        return updatedInstances;
+    }
+
+    private List<String> getIds (List<Instance> instances) {
+        return instances.stream().map(Instance::getInstanceId).collect(Collectors.toList());
+    }
+
+    private String constructUserData(boolean r5) {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+            DescribeInstancesRequest describeInstancesRequest = new DescribeInstancesRequest().withInstanceIds(instanceIds);
+            List<Reservation> reservations = ec2.describeInstances(describeInstancesRequest).getReservations();
+            for (Reservation reservation  : reservations) {
+                for (Instance instance : reservation.getInstances()) {
+                    instanceIpAddresses.put(instance.getInstanceId(), instance.getPublicIpAddress());
+                    updatedInstances.add(instance);
+                }
+            }
+            try {
+                Thread.sleep(10000);
```
evansiroky: Would be nice to have a log statement here to indicate that the program is still running.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+        lines.add(String.format("mkdir -p %s", jarDir));
+        lines.add(String.format("wget %s -O %s/%s.jar", s3JarUrl, jarDir, jarName));
+        if (graphAlreadyBuilt) {
+            lines.add("echo 'downloading graph from s3'");
+            // Download Graph from S3 and spin up trip planner.
+            lines.add(String.format("aws s3 --region us-east-1 cp %s %s/Graph.obj", getS3GraphPath(), routerDir));
+        } else {
+            lines.add("echo 'starting graph build'");
+            // Build the graph if Graph object (presumably this is the first instance to be started up).
+            if (deployment.r5) lines.add(String.format("sudo -H -u ubuntu java -Xmx6G -jar %s/%s.jar point --build %s", jarDir, jarName, routerDir));
```
evansiroky: If we're using this here instead of this method's argument, maybe this method doesn't need an argument.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +409,242 @@ public void jobFinished () {
...
+            else lines.add(String.format("sudo -H -u ubuntu java -jar %s/%s.jar --build %s > $BUILDLOGFILE 2>&1", jarDir, jarName, routerDir));
+            // Upload the graph to S3.
+            if (!deployment.r5) lines.add(String.format("aws s3 --region us-east-1 cp %s/Graph.obj %s", routerDir, getS3GraphPath()));
+        }
+        if (deployment.buildGraphOnly) {
+            lines.add("echo 'shutting down server (build graph only specified in deployment target)'");
```
evansiroky: This would also be a good thing to note somewhere in user documentation... that if only building a graph, the graph building instance will be terminated upon completion and upload of the graph object.

configurations/default/server.yml.tmp
landonreed: Some of this config info might actually make more sense to store in the `OtpServer` object... that might make more sense than splitting up the AWS info by putting the `targetGroupArn` in the `OtpServer` and this stuff in `server.yml`.

landonreed: Also, there may be more than one EC2 environment we're deploying to, so that just strengthens the argument to place this info into `OtpServer`.

evansiroky: I concur.

src/main/java/com/conveyal/datatools/editor/datastore/FeedTx.java
```diff
@@ -124,6 +122,10 @@ public FeedTx(DB tx, boolean buildSecondaryIndices) {
 //        editedSinceSnapshot = tx.getAtomicBoolean("editedSinceSnapshot") == null ? tx.createAtomicBoolean("editedSinceSnapshot", false) : editedSinceSnapshot;
     }
 
+    private static int toGtfsDate (LocalDate date) {
+        return date.getYear() * 10000 + date.getMonthValue() * 100 + date.getDayOfMonth();
```
evansiroky: Needs JavaDoc. Also, is there a way to use a date formatting library here?

src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
```diff
@@ -241,6 +251,33 @@ private static Object updateDeployment (Request req, Response res) {
...
+    }
+
+    /**
+     * Fetches list of {@link EC2InstanceSummary} for all instances matching the provided filters.
+     */
+    public static List<EC2InstanceSummary> fetchEC2InstanceSummaries(Filter... filters) {
+        return fetchEC2Instances(filters).stream().map(EC2InstanceSummary::new).collect(Collectors.toList());
+    }
+
+    public static List<Instance> fetchEC2Instances(Filter... filters) {
```
evansiroky: Needs JavaDoc

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,343 @@
...
+    private static final ObjectMapper mapper = new ObjectMapper();
+    private static final AmazonEC2 ec2 = AmazonEC2Client.builder().build();
+    private static final AmazonIdentityManagement iam = AmazonIdentityManagementClientBuilder.defaultClient();
+    private static final AmazonElasticLoadBalancing elb = AmazonElasticLoadBalancingClient.builder().build();
+
+    /**
+     * Gets the server specified by the request's id parameter and ensure that user has access to the
+     * deployment. If the user does not have permission the Spark request is halted with an error.
+     */
+    private static OtpServer checkServerPermissions(Request req, Response res) {
```
evansiroky: I think this method should be renamed to `getServerWithPermissions` since it returns something.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+        boolean isProjectAdmin = userProfile.canAdministerProject(server.projectId, server.organizationId());
+        if (!isProjectAdmin && !userProfile.getUser_id().equals(server.user())) {
+            // If user is not a project admin and did not create the deployment, access to the deployment is denied.
+            logMessageAndHalt(req, HttpStatus.UNAUTHORIZED_401, "User not authorized for deployment.");
+        }
+        return server;
+    }
+
+    private static OtpServer deleteServer(Request req, Response res) {
```
evansiroky: Needs JavaDoc.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,343 @@
...
+                    logMessageAndHalt(req, 400, message, e);
+                }
+            }
+        } catch (Exception e) {
+            if (e instanceof HaltException) throw e;
+            else logMessageAndHalt(req, 400, "Error encountered while validating server field", e);
+        }
+    }
+
+    private static void validateAmiId(String amiId, Request req) {
```
evansiroky: All these validate messages should have JavaDoc.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    public static boolean amiExists(String amiId) {
+        try {
+            DescribeImagesRequest request = new DescribeImagesRequest().withImageIds(amiId);
+            DescribeImagesResult result = ec2.describeImages(request);
+            // Iterate over AMIs to find a matching ID.
+            for (Image image : result.getImages()) if (image.getImageId().equals(amiId)) return true;
+        } catch (AmazonEC2Exception e) {
+            LOG.info("AMI does not exist.", e);
```
evansiroky: Could it be possible that an error prevented the proper checking of the AMI? I think the error message should be changed to reflect that possibility.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -0,0 +1,343 @@
...
+            // Iterate over subnets. If a matching ID is found, silently return.
+            for (Subnet subnet : result.getSubnets()) if (subnet.getSubnetId().equals(subnetId)) return;
+        } catch (AmazonEC2Exception e) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, message, e);
+        }
+    }
+
+    private static void validateInstanceType(String instanceType, Request req) {
+        if (instanceType == null) return;
+            try {
```
evansiroky: please correct indentation

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
         NotifyUsersForSubscriptionJob.createNotification("deployment-updated", deployment.id, message);
     }
 
+
+    private void replaceEC2Servers() {
```
evansiroky: Needs JavaDoc.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+                    service.submit(monitorServerStatusJob);
+                }
+                // Shutdown thread pool once the jobs are completed and wait for its termination. Once terminated, we can
+                // consider the servers up and running (or they have failed to initialize properly).
+                service.shutdown();
+                service.awaitTermination(4, TimeUnit.HOURS);
+            }
+            String finalMessage = "Server setup is complete!";
+            // Get EC2 servers running that are associated with this server.
+            List<String> instanceIds = previousInstances.stream()
```
evansiroky: I think it'd be good to have a variable name that continues to indicate that these are previous instances. Maybe `previousInstanceIds`.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+            List<Reservation> reservations = ec2.describeInstances(request).getReservations();
+            for (Reservation reservation  : reservations) {
+                for (Instance instance : reservation.getInstances()) {
+                    instanceIpAddresses.put(instance.getInstanceId(), instance.getPublicIpAddress());
+                    updatedInstances.add(instance);
+                }
+            }
+            try {
+                int sleepTimeMillis = 10000;
+                LOG.info("Waiting {} seconds...", sleepTimeMillis / 1000);
```
evansiroky: should say what is being waited for.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+                    .withTags(new Tag("Name", serverName))
+                    .withTags(new Tag("projectId", deployment.projectId))
+                    .withTags(new Tag("deploymentId", deployment.id))
+                    .withTags(new Tag("jobId", this.jobId))
+                    .withTags(new Tag("serverId", otpServer.id))
+                    .withResources(instance.getInstanceId())
+            );
+        }
+        List<Instance> updatedInstances = new ArrayList<>();
+        while (instanceIpAddresses.values().contains(null)) {
```
evansiroky: There is no other escape from this loop in the event that some instances don't get assigned public IPs. This is probably unlikely to happen, but it might be a good idea to add some absurd time limit that fails the deploy job just in case.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+                ExecutorService service = Executors.newFixedThreadPool(remainingServerCount);
+                for (Instance instance : remainingInstances) {
+                    // Note: new instances are added
+                    MonitorServerStatusJob monitorServerStatusJob = new MonitorServerStatusJob(owner, deployment, instance, otpServer);
+                    service.submit(monitorServerStatusJob);
+                }
+                // Shutdown thread pool once the jobs are completed and wait for its termination. Once terminated, we can
+                // consider the servers up and running (or they have failed to initialize properly).
+                service.shutdown();
+                service.awaitTermination(4, TimeUnit.HOURS);
```
evansiroky: Is there some kind of way to update the DeployJob % completion as various MonitorServerStatusJobs complete? Not essential, but would be nice to show that progress.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+        if (deployment.buildGraphOnly) {
+            lines.add("echo 'shutting down server (build graph only specified in deployment target)'");
+            lines.add("sudo poweroff");
+        } else {
+            lines.add("echo 'kicking off trip planner (logs at $LOGFILE)'");
+            // Kick off the application.
+            if (deployment.r5) lines.add(String.format("sudo -H -u ubuntu nohup java -Xmx6G -Djava.util.Arrays.useLegacyMergeSort=true -jar %s/%s.jar point --isochrones %s > /var/log/r5.out 2>&1&", jarDir, jarName, routerDir));
+            else lines.add(String.format("sudo -H -u ubuntu nohup java -jar %s/%s.jar --server --bindAddress 127.0.0.1 --router default > $LOGFILE 2>&1 &", jarDir, jarName));
+        }
+        return String.join("\n", lines);
```
evansiroky: Basically all of these statements are susceptible to failure and don't have an easier way to recover. Also, there is no real way to check on the progress of everything. Maybe this script could also be updating some file that could be on s3 that datatools-server could be polling to see how much progress has been made and whether or not a failure occurred. Otherwise, if an error occurs the user won't know what was the issue and will have to wait an hour for the job to be failed.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+        lines.add(String.format("aws s3 --region us-east-1 cp %s /tmp/bundle.zip", s3BundlePath));
+        lines.add(String.format("unzip /tmp/bundle.zip -d %s", routerDir));
+        // FIXME: Add ability to fetch custom bikeshare.xml file (CarFreeAtoZ)
+        if (false) {
+            lines.add(String.format("wget -O %s/bikeshare.xml ${config.bikeshareFeed}", routerDir));
+            lines.add(String.format("printf \"{\\n  bikeRentalFile: \"bikeshare.xml\"\\n}\" >> %s/build-config.json\"", routerDir));
+        }
+        // Download trip planner JAR.
+        lines.add(String.format("mkdir -p %s", jarDir));
+        lines.add(String.format("wget %s -O %s/%s.jar", s3JarUrl, jarDir, jarName));
```
evansiroky: This could fail and result in the rest of the items failing.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+        if (graphAlreadyBuilt) {
+            lines.add("echo 'downloading graph from s3'");
+            // Download Graph from S3 and spin up trip planner.
+            lines.add(String.format("aws s3 --region us-east-1 cp %s %s/Graph.obj", getS3GraphPath(), routerDir));
```
evansiroky: Needs fault tolerance in case this request fails.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+        } else {
+            lines.add("echo 'starting graph build'");
+            // Build the graph if Graph object (presumably this is the first instance to be started up).
+            if (deployment.r5) lines.add(String.format("sudo -H -u ubuntu java -Xmx6G -jar %s/%s.jar point --build %s", jarDir, jarName, routerDir));
+            else lines.add(String.format("sudo -H -u ubuntu java -jar %s/%s.jar --build %s > $BUILDLOGFILE 2>&1", jarDir, jarName, routerDir));
```
evansiroky: Needs fault tolerance in case graph build fails.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -330,6 +417,258 @@ public void jobFinished () {
...
+            if (!deployment.r5) lines.add(String.format("aws s3 --region us-east-1 cp %s/Graph.obj %s", routerDir, getS3GraphPath()));
+        }
+        if (deployment.buildGraphOnly) {
+            lines.add("echo 'shutting down server (build graph only specified in deployment target)'");
+            lines.add("sudo poweroff");
+        } else {
+            lines.add("echo 'kicking off trip planner (logs at $LOGFILE)'");
+            // Kick off the application.
+            if (deployment.r5) lines.add(String.format("sudo -H -u ubuntu nohup java -Xmx6G -Djava.util.Arrays.useLegacyMergeSort=true -jar %s/%s.jar point --isochrones %s > /var/log/r5.out 2>&1&", jarDir, jarName, routerDir));
+            else lines.add(String.format("sudo -H -u ubuntu nohup java -jar %s/%s.jar --server --bindAddress 127.0.0.1 --router default > $LOGFILE 2>&1 &", jarDir, jarName));
```
evansiroky: Needs fault tolerance in case OTP fails to startup.

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -0,0 +1,145 @@
...
+        String otpUrl = String.format("http://%s/otp", instance.getPublicIpAddress());
+        boolean otpIsRunning = false, routerIsAvailable = false;
+        // Progressively check status of OTP server
+        if (deployment.buildGraphOnly) {
+            // FIXME No need to check that OTP is running. Just check to see that the graph is built.
+//            FeedStore.s3Client.doesObjectExist(otpServer.s3Bucket, );
+        }
+        // First, check that OTP has started up.
+        status.update("Instance status is OK. Waiting for OTP...", 20);
+        while (!otpIsRunning) {
```
evansiroky: As noted in https://github.com/ibi-group/datatools-server/pull/225/files#r324391156, there are many things that occur before OTP starts. These items could be reported to better update progress, or if an fatal error occured, the job could be immediately failed instead of waiting for the timeout.

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -0,0 +1,145 @@
...
+        LOG.info(otpMessage);
+        status.update(otpMessage, 30);
+        // Once this is confirmed, check for the existence of the router, which will indicate that the graph build is
+        // complete.
+        String routerUrl = String.format("%s/routers/default", otpUrl);
+        while (!routerIsAvailable) {
+            LOG.info("Checking that router is available (i.e., graph build or read is finished) at {}", routerUrl);
+            // If the request was successful, the graph build is complete!
+            // TODO: Substitute in specific router ID? Or just default to... default.
+            routerIsAvailable = checkForSuccessfulRequest(routerUrl, 5);
```
evansiroky: No timeout in this while loop means this could result in an infinite loop of checking if OTP started up even when it has already failed.

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        status.update(String.format("Graph build completed on server %s (%s).", instance.getInstanceId(), routerUrl), 90);
+        if (otpServer.ec2Info.targetGroupArn != null) {
+            // After the router is available, the EC2 instance can be registered with the load balancer.
+            // REGISTER INSTANCE WITH LOAD BALANCER
+            AmazonElasticLoadBalancing elbClient = AmazonElasticLoadBalancingClient.builder().build();
+            RegisterTargetsRequest registerTargetsRequest = new RegisterTargetsRequest()
+                    .withTargetGroupArn(otpServer.ec2Info.targetGroupArn)
+                    .withTargets(new TargetDescription().withId(instance.getInstanceId()));
+            elbClient.registerTargets(registerTargetsRequest);
```
evansiroky: Could this throw an exception if it fails? If so, we want to be sure to catch that and fail the job accordingly.

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -55,6 +58,9 @@
 
     public String name;
 
+    public static final String DEFAULT_OTP_VERSION = "otp-v1.3.0";
```
evansiroky: 1.4 is the latest. Should we use that?

src/main/java/com/conveyal/datatools/manager/models/EC2Info.java
```diff
@@ -0,0 +1,36 @@
...
+    public String subnetId;
+    /** The security group ID associated with the target group. */
+    public String securityGroupId;
+    /** The Amazon machine image (AMI) to be used for the OTP EC2 machines. */
+    public String amiId;
+    /** The IAM role ARN that the OTP EC2 server should assume. */
+    public String iamRoleArn;
+    /** The AWS key file (.pem) that should be used to set up OTP EC2 servers (gives a way for admins to SSH into machine). */
+    public String keyName;
+    /** The target group to deploy new EC2 instances to. */
```
evansiroky: Would be nice to know that this is specifically a Target Group associated with a certain ELB.

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -397,6 +397,8 @@ public void delete() {
                 Persistence.feedSources.update(fs.id, "{lastFetched:null}");
             }
             feedStore.deleteFeed(id);
+            // Delete feed version tables in GTFS database
+            GTFS.delete(this.namespace, DataManager.GTFS_DATA_SOURCE);
```
evansiroky: See https://github.com/conveyal/gtfs-lib/issues/248.

src/test/java/com/conveyal/datatools/DatatoolsTest.java
```diff
@@ -43,7 +43,7 @@ public static void setUp() throws RuntimeException, IOException {
         // Travis CI, these files will automatically be setup.
         String[] args = getBooleanEnvVar("RUN_E2E")
             ? new String[] { "configurations/default/env.yml", "configurations/default/server.yml" }
-            : new String[] { "configurations/default/env.yml.tmp", "configurations/default/server.yml.tmp" };
+            : new String[] { "configurations/test/env.yml.tmp", "configurations/test/server.yml.tmp" };
```
evansiroky: 👍

src/main/java/com/conveyal/datatools/editor/datastore/FeedTx.java
landonreed: This is deprecated code. It honestly probably doesn't even need to exist in this version. I'm going to create an issue to discuss removing this.

landonreed: Issue is #236.

src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
```diff
@@ -84,6 +88,46 @@ private static Deployment deleteDeployment (Request req, Response res) {
...
+            }
+        } else {
+            summaryToDownload = deployment.latest();
+        }
+        if (summaryToDownload == null) {
+            // Try to construct the URI string
+            OtpServer server = Persistence.servers.getById(deployment.deployedTo);
+            if (server == null) {
+                uriString = String.format("s3://%s/bundles/%s/%s/%s", "S3_BUCKET", deployment.projectId, deployment.id, jobId);
+                logMessageAndHalt(req, 400, "Cannot construct URI for build artifact. " + uriString);
```
evansiroky: Seems like this error message could be better worded to indicate that the deployment doesn't have server information.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -404,74 +423,107 @@ public void jobFinished () {
...
+            List<MonitorServerStatusJob> remainingServerMonitorJobs = new ArrayList<>();
+            List<Instance> remainingInstances = new ArrayList<>();
             if (remainingServerCount > 0) {
                 // Spin up remaining EC2 instances.
                 status.message = String.format("Spinning up remaining %d instance(s).", remainingServerCount);
-                List<Instance> remainingInstances = startEC2Instances(remainingServerCount);
-                instances.addAll(remainingInstances);
+                remainingInstances.addAll(startEC2Instances(remainingServerCount, true));
+                if (remainingInstances.size() == 0 || status.error) {
+                    ServerController.terminateInstances(getIds(remainingInstances));
```
evansiroky: It's only 1 line of code, but it may still make sense to create a method to terminate instances given some IDs.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -575,98 +659,155 @@ private void replaceEC2Servers() {
...
+            final URL url = new URL(s3JarUrl);
+            HttpURLConnection huc = (HttpURLConnection) url.openConnection();
+            huc.setRequestMethod("HEAD");
+            int responseCode = huc.getResponseCode();
+            if (responseCode != HttpStatus.OK_200) {
+                status.fail(String.format("Requested trip planner jar does not exist at s3://%s/%s", s3JarBucket, s3JarKey));
+                return null;
+            }
+        } catch (IOException e) {
+            status.fail(String.format("Error checking for trip planner jar: s3://%s/%s", s3JarBucket, s3JarKey));
```
evansiroky: This exception should be logged.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -575,98 +659,155 @@ private void replaceEC2Servers() {
...
+        String s3JarKey = jarName + ".jar";
+        // If jar does not exist in bucket, fail job.
+        String s3JarUrl = String.format("https://%s.s3.amazonaws.com/%s", s3JarBucket, s3JarKey);
+        try {
+            final URL url = new URL(s3JarUrl);
+            HttpURLConnection huc = (HttpURLConnection) url.openConnection();
+            huc.setRequestMethod("HEAD");
+            int responseCode = huc.getResponseCode();
+            if (responseCode != HttpStatus.OK_200) {
+                status.fail(String.format("Requested trip planner jar does not exist at s3://%s/%s", s3JarBucket, s3JarKey));
```
evansiroky: Might be good to have a log statement as well here.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -575,98 +659,155 @@ private void replaceEC2Servers() {
...
-            if (!deployment.r5) lines.add(String.format("aws s3 --region us-east-1 cp %s/Graph.obj %s", routerDir, getS3GraphPath()));
+            // Upload the graph and build log file to S3.
+            if (!deployment.r5) {
+                lines.add(String.format("aws s3 --region us-east-1 cp %s/%s %s ", routerDir, OTP_GRAPH_FILENAME, getS3GraphURI()));
+                String s3BuildLogPath = joinToS3FolderURI(getBuildLogFilename());
+                lines.add(String.format("aws s3 --region us-east-1 cp $BUILDLOGFILE %s ", s3BuildLogPath));
+            }
         }
+        // Upload user data log.
+        lines.add("instance_id=`curl http://169.254.169.254/latest/meta-data/instance-id`");
```
evansiroky: Can you add a comment explaining what this line does?

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -575,98 +659,155 @@ private void replaceEC2Servers() {
...
             // Build the graph if Graph object (presumably this is the first instance to be started up).
             if (deployment.r5) lines.add(String.format("sudo -H -u ubuntu java -Xmx6G -jar %s/%s.jar point --build %s", jarDir, jarName, routerDir));
             else lines.add(String.format("sudo -H -u ubuntu java -jar %s/%s.jar --build %s > $BUILDLOGFILE 2>&1", jarDir, jarName, routerDir));
-            // Upload the graph to S3.
-            if (!deployment.r5) lines.add(String.format("aws s3 --region us-east-1 cp %s/Graph.obj %s", routerDir, getS3GraphPath()));
+            // Upload the graph and build log file to S3.
+            if (!deployment.r5) {
+                lines.add(String.format("aws s3 --region us-east-1 cp %s/%s %s ", routerDir, OTP_GRAPH_FILENAME, getS3GraphURI()));
+                String s3BuildLogPath = joinToS3FolderURI(getBuildLogFilename());
+                lines.add(String.format("aws s3 --region us-east-1 cp $BUILDLOGFILE %s ", s3BuildLogPath));
```
evansiroky: It may make more sense to upload the logs first in case the graph wasn't built. If the graph wasn't built, the script might error out on a file not found error perhaps.

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -94,24 +146,56 @@ public void jobLogic() {
...
         try {
-            Thread.sleep(1000 * delaySeconds);
+            LOG.info("Waiting {} seconds for {}", DELAY_SECONDS, waitingFor);
+            Thread.sleep(1000 * DELAY_SECONDS);
         } catch (InterruptedException e) {
             e.printStackTrace();
         }
+    }
+
+    private String getBundleStatusKey () {
```
evansiroky: Needs javadoc

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
landonreed: Will that actually happen? I suppose I can make the change regardless, but I don't think the execution of any line is dependent on the previous line, is it?

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
evansiroky: Confirmed that this line can throw an exception. I got this error:

```
com.amazonaws.services.elasticloadbalancingv2.model.InvalidTargetException: The following targets are not in a running state and cannot be registered: 'i-0d76a43db722632ca' (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: InvalidTarget; Request ID: 9d7acd6c-1639-4947-8a10-d8c1837d8959)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1712)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1367)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1113)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:770)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:744)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:726)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:686)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:668)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:532)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:512)
	at com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancingClient.doInvoke(AmazonElasticLoadBalancingClient.java:2765)
	at com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancingClient.invoke(AmazonElasticLoadBalancingClient.java:2732)
	at com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancingClient.invoke(AmazonElasticLoadBalancingClient.java:2721)
	at com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancingClient.executeRegisterTargets(AmazonElasticLoadBalancingClient.java:2303)
	at com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancingClient.registerTargets(AmazonElasticLoadBalancingClient.java:2274)
	at com.conveyal.datatools.manager.jobs.MonitorServerStatusJob.jobLogic(MonitorServerStatusJob.java:147)
```

The job was failed and the DeployJob terminated the instance automatically, so there wasn't even a chance to try to debug what happened. The user logs were uploaded, but don't give any clue as to why the instance wasn't in a running state. Very strange.

evansiroky:
See other comments [here](https://github.com/ibi-group/datatools-ui/pull/463#issuecomment-534775283).

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -317,19 +423,399 @@ public void jobFinished () {
...
+     */
+    private String getRouterId() {
+        return deployment.routerId == null ? "default" : deployment.routerId;
+    }
+
+    /**
+     * Construct the user data script (as string) that should be provided to the AMI and executed upon EC2 instance
+     * startup.
+     */
+    private String constructUserData(boolean graphAlreadyBuilt) {
```
evansiroky: I'm increasingly hesitant about the design of this method. I think this is going to limit us moving forward to be writing bash in java strings. Something I would really like to see is more detailed reporting and logging of the progress of building and loading a graph and automatically retrying if something fails that can be retried. We could attempt to do this with bash, but it would be ugly. If we're trying to ship this quick, then fine I can live with it, but I think we should be making a separate startup script specifically for this where we can do more detailed things. Furthermore, right now this is basically using Amazon s3 as a database of sorts for storing various information when we already have a mongo and postgres instance. Maybe it would make sense for the startup script to directly edit the mongo db or maybe make queries to datatools-server. Either way, we should be thinking about this feature long-term and about what the easiest way to create maintainable code will be.

landonreed: Thanks for the comments. Perhaps there is a better way to handle what gets run in user data, but at some level it is going to require running some bash (unless we opt to include some kind of startup script on the AMI or write a script to a file that gets fetched by the instance). Because of this, I think our best option for now is to keep everything contained in a single place -- this bash script.

Regarding using S3 as a database, can you describe what you mean more fully? We're just storing static files that were generated as a result of the deploy process. We also store GTFS zip files in S3 as well. I'm really not too sure of what the concern there is. The alternative you're proposing as I understand it (the OTP instance connecting to the mongo instance and handing off files/messages) sounds far more complex.

evansiroky: That's what I'm saying. I'm recommending creating a script in some kind of programming language - be it java, python or JavaScript. Then this startup script would communicate directly with either mongo db or datatools-server via HTTP requests perhpas. As it is right now, all of this bash is in one place, but we have to construct it using strings in java. Therefore, we can't view it nicely formatted in an IDE, we can't lint or do static typing analysis, it's much harder to import 3rd party libraries and it is impossible to write isolated tests.

Regarding s3 I'm fine with most of what's going on. Most of the data makes sense to store in a raw format (logs, GTFS files, OSM pbf). The thing in particular that I'm concerned about is constantly checking for bundle download and graph build completion and who knows what else we may add in the future by checking for the existence of an s3 file. I'm not recommending that OTP connect to the mongo instance, but I am recommending that a replacement of the bash script do this and handle the checks of bundle download, graph build, and graph load. These intermediate files that are waited upon seem like the more complex option to me when the bash replacement script could just be writing to the mongo database directly or making an HTTP request to datatools-server, or maybe even using some other messaging framework.

Furthermore, using s3 as a database like this will cost more money since s3 has per-request pricing so the more we scale the more this will cost us. The cost seems lowish at $0.005 per 1k requests, but let's say new graphs are built on a nightly basis. And let's say graph building takes 1 hour for a largeish region with elevation data. So, it'll make an s3 request every 5 seconds for an hour or 3,600 requests per day and 1.3 million a year. If we have 100 regions doing this, the cost of just these s3 requests would be $657 a year compared to $0 for using a database or sending over an HTTP request to the datatools-server instance.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -121,6 +130,16 @@ public String getDeploymentId () {
         return deployment.id;
     }
 
+    /** Increment the completed servers count (for use during ELB deployment) and update the job status. */
+    public void incrementCompletedServers() {
```
evansiroky: drake yes

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -88,52 +92,70 @@ public void jobLogic() {
             }
             // First, check that OTP has started up.
             status.update("Prepping for graph build...", 20);
+            String bundleUrl = String.join("/", ipUrl, BUNDLE_DOWNLOAD_COMPLETE_FILE);
+            long bundleDownloadStartTime = System.currentTimeMillis();
```
evansiroky: I like this addition as well.

PR #234, opened 2019-09-17
fix(merge): omit agency_id field from routes in mtc merges

### LINKED ISSUE #233: agency_id should not be added to routes table in MTC merges
## Observed behavior (please include a screenshot if possible)

During feed merges, the _agency_id_ field is added to the _routes_ table even when it's not necessary. This breaks MTC's workflow.

## Expected behavior

During MTC feed merges, the _agency_id_ field should never be added to the _routes_ table.

Attempt to merge 2 fields.

MTC extension is enabled in server.yml file.

## Version of datatools-ui and datatools-server if applicable (exact commit hash or branch name)

https://github.com/ibi-group/datatools-server/commit/cf73a2f2e017a09720059bd2b5af095f479c69da

https://github.com/conveyal/datatools-ui/commit/4af00b7c61eb66731c22f1b3d7bbc8c30d5da695

evansiroky:
### Checklist

### Description

From the email forwarded to me, it seems that MTC wants to never have the _agency_id_ field included in the _routes.txt_ file. So I made some code that does just that. 

Fixes #233.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/MergeFeedsJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/MergeFeedsJob.java
@@ -333,19 +333,15 @@ public class MergeFeedsJob extends MonitorableJob {
         int mergedLineNumber = 0;
         // Get the spec fields to export
         List<Field> specFields = table.specFields();
+        // remove some fields that MTC doesn't want
         if (DataManager.isExtensionEnabled("mtc")) {
             // Remove route and agency branding URL from field list.
-            // Nothing to do for other tables.
             if ("agency".equals(table.name) || "routes".equals(table.name)) {
-                int indexToRemove = -1;
-                for (int i = 0; i < specFields.size(); i++) {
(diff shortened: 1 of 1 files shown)


landonreed:
Waiting for confirmation from MTC that `agency_id` should *never* be included in an exported feed. I somewhat doubt that this was their intent.

evansiroky:
The latest refactor should address all of these use cases.

landonreed:
Sorry to go back and forth on this. I think a more general way to handle this might be to "peek" at each file to be merged and get the common fields shared between them. Then, instead of iterating over all of the specFields, we could iterate over the sharedFields and safely write out all of these fields. I tried this out here: #235

evansiroky:
While #235 seems like a reasonable thing to do, I don't believe #235 would account for scenario 3:

> Scenario 3: agency_id field is included in routes.txt file of both active and future feeds, but without values. Merged feed should eliminate the field.

landonreed:
Ah, you're right. Checking to see if they can live without that case.

landonreed:
Closing in favor of fixes contained in #235.

Issue #156, opened 2018-12-20
Type safety issue when updating application data stored in MongoDB


landonreed:
There is a weakness in how Data Tools server handles updates to application data, which has been lingering in the Mongo-backed code for some time.

Essentially, in many of the update HTTP handlers, we take a JSON document and write it directly into the Mongo database. There are no type checks and an update with a bad JSON document could basically break the codecs that handle deserializing the data back into POJOs. I was reminded of this recently when @evansiroky was working on some updates in a recent PR (#151).

One obvious way is to just read the JSON into POJOs first and then use Mongo’s `replaceOne` method to overwrite the object with the updated data. There are places in the application where we have requests to update just a single field on an object (for example, updating a feed version’s name), so we would have to consider if we need any separate HTTP handlers to handle these requests.

landonreed:
I think my preferred approach would be to add a function to `SparkUtils` that looks like the following:

```
public static <T> T readObjectFromRequest (Request req, Class<T> clazz) {
        try {
            return mapper.readValue(req.body(), clazz);
        } catch (IOException e) {
            haltWithMessage(req, 400, String.format("Could not read %s object from JSON request body.", clazz.getSimpleName()));
            return null;
        }
    }
```

We do also need to go through all of the update controllers and UI code to make sure we take account of any places where partial updates are being made (e.g., updating just one field). We'll need a separate type-safe strategy for that (I'm open to suggestions).

PR #249, opened 2019-10-17
Fix FeedSource update, misc cleanup

landonreed:
### Checklist

### Description

fix ibi-group/datatools-ui#502, ref #156


src/main/java/com/conveyal/datatools/manager/auth/Auth0UserProfile.java
```diff
@@ -346,6 +347,23 @@ public boolean canAdministerProject(String projectID, String organizationId) {
         return false;
     }
 
+    /** Check that user can administer project. Organization ID is drawn from persisted project. */
+    public boolean canAdministerProject(String projectId) {
+        if (canAdministerApplication()) return true;
+        com.conveyal.datatools.manager.models.Project p = Persistence.projects.getById(projectId);
```
evansiroky: Is there a way to not have the full classpath here?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -40,130 +41,118 @@
...
-            }
-        } else if (userId != null) {
-            // request feed sources a specified user has permissions for
-            Auth0UserProfile user = getUserById(userId);
-            if (user == null) return feedSourcesToReturn;
-
-            for (FeedSource source: Persistence.feedSources.getAll()) {
+        // Determine whether request is coming from a public HTTP endpoint.
+        boolean publicFilter = req.pathInfo().contains("public");
+        List<FeedSource> allFeedSources = Persistence.feedSources.getAll();
```
evansiroky: I'm guessing we're not going to have that many feedSources to search through so it's fine to keep this for now, but it seems like there could be a way to query the Persistence to get the desired feed sources.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -40,130 +41,118 @@
...
                 String orgId = source.organizationId();
                 if (
-                    source != null && source.projectId != null &&
-                    (user.canManageFeed(orgId, source.projectId, source.id) || user.canViewFeed(orgId, source.projectId, source.id))
+                    source.projectId != null && source.projectId.equals(projectId) &&
+                    requestingUser.canManageOrViewFeed(orgId, source.projectId, source.id)
                 ) {
-
+                    // If requesting public sources and source is not public; skip source
+                    // TODO: determine if this is still needed.
```
evansiroky: Maybe we should determine this now. Didn't we already get rid of a bunch of public endpoints?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -40,130 +41,118 @@
...
+            // created in our database.
+            for (String resourceType : DataManager.feedResources.keySet()) {
+                DataManager.feedResources.get(resourceType).feedSourceCreated(newFeedSource, req.headers("Authorization"));
+            }
+            // Notify project subscribers of new feed source creation.
+            Project parentProject = Persistence.projects.getById(newFeedSource.projectId);
+            NotifyUsersForSubscriptionJob.createNotification(
+                "project-updated",
+                newFeedSource.projectId,
+                String.format("New feed %s created in project %s.", newFeedSource.name, parentProject.name));
```
evansiroky: I prefer having the ending paren be on a newline.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -40,130 +41,118 @@
...
+            return null;
+        }
     }
 
-    public static FeedSource updateFeedSource(Request req, Response res) {
+    /**
+     * Spark HTTP endpoint to update a feed source. Note: at one point this endpoint accepted a JSON object
+     * representing a single field to update for the feed source, but it now requires that the JSON body represent all
+     * fields the updated feed source should contain. This change allows us to parse the JSON into a POJO, which
+     * essentially does type checking for us and prevents issues with deserialization from MongoDB into POJOs.
```
evansiroky: I prefer this method and think it'll make maintenance easier.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -40,130 +41,118 @@
...
-            // Reset last fetched timestamp if the URL has been updated.
-            if (!fieldsToUpdate.get("url").toString().equals(formerFeedSource.url.toString())) {
-                LOG.info("Feed source fetch URL has been modified. Resetting lastFetched value from {} to {}", formerFeedSource.lastFetched, null);
-                fieldsToUpdate.put("lastFetched", null);
-            }
+        FeedSource updatedFeedSource = getPOJOFromRequestBody(req, FeedSource.class);
+        // Feed source previously had a URL, but it has been changed. In this case, we reset the last fetched timestamp.
+        if (formerFeedSource.url != null && !formerFeedSource.url.equals(updatedFeedSource.url)) {
+            LOG.info("Feed source fetch URL has been modified. Resetting lastFetched value from {} to {}", formerFeedSource.lastFetched, null);
+            updatedFeedSource.lastFetched = null;
```
evansiroky: Are we sure we need to do this? Seems like we'd lose valuable information about when a feed was last fetched even if it was from an old url.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -234,11 +222,12 @@ private static FeedSource deleteFeedSource(Request req, Response res) {
     /**
      * Re-fetch this feed from the feed source URL.
      */
-    public static String fetch (Request req, Response res) {
+    private static String fetch (Request req, Response res) {
         FeedSource s = requestFeedSourceById(req, Actions.MANAGE);
-
-        LOG.info("Fetching feed for source {}", s.name);
-
+        if (s.url == null) {
```
evansiroky: Might be good to add a check for whether the URL is valid or not before initiating the fetchSingleFeedJob.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -264,12 +253,14 @@ public static FeedSource requestFeedSourceById(Request req, Actions action) {
...
         Auth0UserProfile userProfile = req.attribute("user");
-        Boolean publicFilter = Boolean.valueOf(req.queryParams("public")) ||
+        boolean publicFilter = Boolean.valueOf(req.queryParams("public")) ||
                 req.url().split("/api/*/")[1].startsWith("public");
 
         // check for null feedSource
-        if (feedSource == null)
+        if (feedSource == null) {
             logMessageAndHalt(req, 400, "Feed source ID does not exist");
+            return null;
```
evansiroky: Do we need to return null after a `logMessageAndHalt` now?

src/main/java/com/conveyal/datatools/manager/auth/Auth0UserProfile.java
landonreed: Not that I'm aware of.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: Indeed, good point. But per your other good point, we might just need to remove this because most traces of the public API have been ripped out already.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: Yea good point. I'll do a double check, but then just go ahead and remove.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: Do you have an alternative approach in mind, @evansiroky?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: Well, it silences warnings from Intellij about objects (e.g., `feedSource` in this case) being null farther down in the method body. Maybe a better approach would to give `logMessageAndHalt` a return value of null and add a `return` in front of anytime we call that method?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: I think the `URL` constructor will instantiate as null (and throw exceptions) if the URL is deemed to be invalid.

landonreed:
Questions responded to. Will work on changes now.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Yes, it will do that, but it'll be inside the job instead of the HTTP request, so there will be a bit of a delay for the user. It might be a better user experience to make the HTTP request fail so the user sees the error sooner.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Seems like it'd make sense to store and retrieve that information from the most recent Feed Version. That might require adding additional information to the Feed Version model though.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Oh interesting, I'm not getting those warnings in IntelliJ, but I could see that being useful. How did you enable these warnings?

evansiroky:
Assigning back now that some questions have been answered.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: What I'm suggesting is that this conditional `s.url == null` will catch the case where the URL field is invalid. When the feed source is drawn from MongoDB, the URL codec will fail to instantiate the URL value (I think) and it will appear as null in the FeedSource object field here. I could be wrong, but that's my understanding.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: This is not really new code, so if you think it's worth pulling a ticket for to track and continue the conversation, feel free. I'm not sure we should make these changes in this context though.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: Not exactly sure haha, but it's called "method invocation may produce nullpointerexception."

PR #253, opened 2019-10-18
MTC service_id merge update

### LINKED ISSUE #244: MTC feed merge: feed-scope service_ids duplicated in both calendar_dates tables
From MTC:
>we would like to update all the calender_dates.txt records and their corresponding records in trips,txt file in the active feed. This helps us to resolve a conflict between service_ids with same name in active and future feeds. Please let us know if you have any questions or concerns regarding this request.
Calendar_dates updates_TDM

Essentially, what we need to do here is check if any service_id shows up in both `calendar_dates.txt` files. If it does, we should feed-scope the service_id for one of those feeds. We already do this for duplicate service_ids found between the `calendars.txt` tables, so any service_id that is in both feeds' calendars and calendar_dates will get feed-scoped. Operationally, we probably just need to handle this within constructMergedTable, keeping a record of all service_ids from calendar_dates and feed-scope any dupes. There might be some tricky edge cases here. For example, let's say service_id `ABC` exists in feed 1 calendars.txt but not in feed 2 calendars.txt, but it exists in both calendar_dates.txt files. We may need to ensure that when writing out the GTFS that if `ABC` is feed-scoped in feed 1 (e.g., `VTA:ABC`) that it gets properly written out/feed scoped in calendars.txt. calendars.txt gets processed before calendar_dates, so that might complicate things a bit.

GTFS Feeds:
[VTA-Future-Feed-10-3-2019.zip](https://github.com/ibi-group/datatools-server/files/3709056/VTA-Future-Feed-10-3-2019.zip)
[VTA-Current-Feed_10-3-2019.zip](https://github.com/ibi-group/datatools-server/files/3709057/VTA-Current-Feed_10-3-2019.zip)

evansiroky:
### Checklist

### Description

Resolves #244 by always feed-scoping the `service_id`s encountered in the chronologically first feed version that is merged.

Also:
- adds test for service_id feed scoping during MTC merges
- consolidates some test code
- moves all GTFS files/folders used for testing to the `src/test/resources/com/conveyal/datatools/gtfs` folder.


src/main/java/com/conveyal/datatools/manager/jobs/MergeFeedsJob.java
```diff
@@ -518,9 +518,20 @@ private int constructMergedTable(Table table, List<FeedToMerge> feedsToMerge,
...
                         // York State).
                         if (mergeType.equals(MTC)) {
-                            Set<NewGTFSError> idErrors = referenceTracker
-                                .checkReferencesAndUniqueness(keyValue, lineNumber, field, val,
-                                    table, keyField, orderField);
+                            Set<NewGTFSError> idErrors;
+                            // If analyzing the second feed (non-future feed), the service_id always gets feed scoped.
+                            // See https://github.com/ibi-group/datatools-server/issues/244
+                            if (feedIndex == 1 && field.name.equals("service_id")) {
+                                valueToWrite = String.join(":", idScope, val);
```
landonreed: I think this feed-scoped value needs to get updated in mergedFeedsResult:
```
mergeFeedsResult.remappedIds.put(key, valueToWrite);
```

evansiroky: Hmmm, now that you mention that, it looks like there could be a bug that happens on remapping calendar IDs of a feed that only has calendar_dates. I'm going to make a test case and fix for that.

evansiroky: Well, after quite a bit of investigation and more testing I did add it the code you mentioned, but the effect of doing that is nearly inconsequential when there is a feed that only has calendar_dates. I wrote a test case for this and since my code will scope any service_id for any field, this additional check and renaming doesn't happen for calendar_dates. Regardless, the resulting merged feed is technically correct and passes the test case. However, the MergeFeedResult#remappedReferences is not incremented properly when the scoped feed has only calendar_dates. That metric is displayed in the UI, but I'm wondering do we really need to show it?

Additionally, I noticed what might be 2 other problems:

1. Odd skipping logic

In the calculations for whether a service_id in an active feed should be skipped, the comparison of this is done against the current day, not the start day of the future feed. This seems odd. It seems like it would make more sense to have the comparison date be the start date of the future feed and then skip various service_id's in the active feed if they occur after that date.

2. calendar_dates skipping

There is currently no logic for skipping service_ids that occur in the calendar_dates file. Shouldn't there also be logic in this area?

evansiroky:
See additional comments and questions in https://github.com/ibi-group/datatools-server/pull/253#discussion_r338886946.

src/main/java/com/conveyal/datatools/manager/jobs/MergeFeedsJob.java
landonreed: Waiting on response for item 2 from MTC. Both items currently captured in #256.

landonreed:
@evansiroky, following up on this PR so we can get it merged in. As you are aware, MTC has elected to move forward with the calendar_dates check, but let's handle that in a separate PR. We'll also need to address the other part of #256 in a separate PR (I need to check with MTC about that item as well).

In the meantime, could you fix the conflicts and re-assign to me once it's ready for a final check?

evansiroky:
I fixed the merge conflicts.

landonreed:
Not sure what happened, but GitHub just decided there were more conflicts. Can you investigate, @evansiroky?

landonreed:
Never mind, I think GitHub just hiccuped. I pulled the branches locally, did a merge, and git just did a fast forward.

PR #254, opened 2019-10-21
Add API route for resending email verifcation

evansiroky:
### Checklist

### Description

Adds the necessary backend API endpoint to allow initiating the sending of a user verification email.

See companion PR in the UI repo: https://github.com/ibi-group/datatools-ui/pull/505


src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -300,6 +304,45 @@ private static Object getRecentActivity(Request req, Response res) {
...
+            );
+
+            // Execute request. If a HTTP response other than 200 occurs, an error will be returned.
+            executeRequestAndGetResult(resendEmailVerificationRequest, req);
+
+            // Email verification successfully sent! Return successful response.
+            return JsonUtil.objectMapper.createObjectNode()
+                .put("emailSent", true);
+        } else {
+            logMessageAndHalt(
```
landonreed: Can this else block get a comment about conditions that might lead to an unauthorized response?

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -300,6 +304,45 @@ private static Object getRecentActivity(Request req, Response res) {
...
     }
 
+    /**
+     * Resends the user confirmation email for a given user
+     */
+    private static ObjectNode resendEmailConfirmation(Request req, Response res) {
+        // verify if the request is legit. Should only come from an application administrator or the user for which the
+        // account belongs to.
+        Auth0UserProfile userProfile = req.attribute("user");
+        if (userProfile.canAdministerApplication() || userProfile.getUser_id().equals(req.params("id"))) {
```
landonreed: When would this request be coming from an application administrator? Perhaps this needs a TODO describing that this never actually happens in the current UI, but is some kind of future proofing? It might should just be removed though. Curious to hear your thoughts.

evansiroky: Yeah, it's not currently used anywhere in the current app, but it sort of seemed like something a thing that the app admin would be able to do. I'll just remove it and we can add it back later if needed.

evansiroky:
Just refactored to address comments.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
landonreed: This comment is out of date now.

PR #296, opened 2020-04-09
Fail a deploy job on unhealthy instance state

### LINKED ISSUE #295: If a user terminates a graph building instance before it completes, the deploy job will hang for hours.
## Observed behavior

If a user terminates a graph building instance before it completes, the deploy job will hang for hours.

## Expected behavior

Datatools server should automatically fail a deploy job if the graph building instance state is terminated by the user before graph building completes.

Start a deployment, then terminate the graph building instance. The deploy job will continue until it times out.

With ec2 deployments enabled.

fd738dd797dda97529a399811598a34d6dd75437

evansiroky:
### Checklist

### Description

Fixes #295 by automatically terminating graph building instances if they have been manually terminated by the user.

### DIFF
--- a/src/main/java/com/conveyal/datatools/common/status/MonitorableJob.java
+++ b/src/main/java/com/conveyal/datatools/common/status/MonitorableJob.java
@@ -179,11 +179,9 @@ public abstract class MonitorableJob implements Runnable, Serializable {
             // Run final steps of job pending completion or error. Note: any tasks that depend on job success should
             // check job status to determine if final step should be executed (e.g., storing feed version in MongoDB).
             // TODO: should we add separate hooks depending on state of job/sub-tasks (e.g., success, catch, finally)
-            // Set status to finished in case duration needed by finishing stage. (e.g., storing the job duration in a
-            // database).
-            if (cancelMessage.equals("")) {
-                status.completeSuccessfully("Job complete!");
-            }
+            // Complete the job (as success if no errors encountered, as failure otherwise).
+            if (!parentJobErrored && !subTaskErrored) status.completeSuccessfully("Job complete!");
+            else status.complete(true);
             jobFinished();
 
             // We retain finished or errored jobs on the server until they are fetched via the API, which implies they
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/datatools/common/status/MonitorableJob.java
```diff
@@ -181,7 +181,9 @@ public void run () {
             // TODO: should we add separate hooks depending on state of job/sub-tasks (e.g., success, catch, finally)
             // Set status to finished in case duration needed by finishing stage. (e.g., storing the job duration in a
             // database).
-            status.completeSuccessfully("Job complete!");
+            if (cancelMessage.equals("")) {
+                status.completeSuccessfully("Job complete!");
+            }
```
landonreed, line 186: Could you update the comment for this conditional? Why does an empty string indicate success... I feel like this would be a little more legible if you were checking that either the subtask or parent task errored.

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -87,131 +91,135 @@ public void jobLogic() {
...
         boolean routerIsAvailable = false, graphIsAvailable = false;
         // If graph was not already built by a previous server, wait for it to build.
-        if (!graphAlreadyBuilt) {
-            boolean bundleIsDownloaded = false;
-            // Progressively check status of OTP server
-            if (deployment.buildGraphOnly) {
-                // No need to check that OTP is running. Just check to see that the graph is built.
-                bundleIsDownloaded = true;
-                routerIsAvailable = true;
+        try {
```
landonreed, line 94: Does this whole block really need to be wrapped in the try/catch? Could we localize this a little more to where the instance health check happens?

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -87,131 +91,135 @@ public void jobLogic() {
...
-                String.format(
-                    "Server successfully registered with load balancer %s. OTP running at %s",
-                    otpServer.ec2Info.targetGroupArn,
-                    routerUrl
-                )
-            );
-            LOG.info("View logs at {}", getUserDataLogS3Path());
-            deployJob.incrementCompletedServers();
-        } else {
-            failJob("There is no load balancer under which to register ec2 instance.");
```
landonreed, line 214: This else block should probably just move to the very top of the jobLogic. We should fail immediately before we even start up a server, right?

src/main/java/com/conveyal/datatools/common/status/MonitorableJob.java
landonreed: I'm going to go ahead and implement this fix.

PR #297, opened 2020-04-10
Add call to MTCValidator

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected (to mtc-20200310)

### Description

This PR adds a call to `MTCValidator` and updates gtfsplus.yml to match MTC guidelines.

`MTCValidator` is available on the [field-lengh-validator (MTC) branch of gtfs-lib](https://github.com/conveyal/gtfs-lib/tree/field-length-validator) or on the main branch if (https://github.com/conveyal/gtfs-lib/pull/276 is merged (the normal CI process fails the build because of that).

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
@@ -33,6 +33,7 @@ import java.util.Date;
 import java.util.Set;
 import java.util.UUID;
 
+import static com.conveyal.datatools.manager.DataManager.isExtensionEnabled;
 import static com.conveyal.datatools.manager.utils.StringUtils.getCleanName;
 import static com.mongodb.client.model.Filters.and;
 import static com.mongodb.client.model.Filters.eq;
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -313,11 +310,9 @@ public void validate(MonitorableJob.Status status) {
             // run validation on feed version
             // FIXME: pass status to validate? Or somehow listen to events?
             status.update("Validating feed...", 33);
-            validationResult = GTFS.validate(feedLoadResult.uniqueIdentifier, DataManager.GTFS_DATA_SOURCE);
+            validationResult = GTFS.validate(feedLoadResult.uniqueIdentifier, DataManager.GTFS_DATA_SOURCE, MTCValidator::new);
```
landonreed, line 313: @binh-dam-ibigroup, I think we should only run the MTCValidator if the MTC extension is enabled `isExtensionEnabled("mtc")`. We only want the MTC-specific checks running on MTC instances of Data Tools, not for everyone.

landonreed (changes requested):
One change needed with respect to the MTCValidator. Also, the gtfs-lib dependency is ready to be updated in `pom.xml`.

landonreed (approved):

PR #309, opened 2020-05-05
GTFS transform

landonreed:
### Checklist

### Description

This adds the ability to define a list of arbitrary transformations that should be applied to each
new feed version loaded for a particular feed source. The primary use case for now is limited to
adding/replacing a file (in this case a GTFS+) file into each new version.

There is not a corresponding UI PR yet, because I would like to stabilize this before we build out a fancy UI to match ✨. There are, however, some tests you can run in `ArbitraryFeedTransformJobTest.java` (if anyone thinks this name is too cumbersome, please let me know).

re ibi-group/datatools-ui#544


src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -113,6 +118,23 @@ private static FeedSource createFeedSource(Request req, Response res) throws IOE
         }
     }
 
+    private static boolean validate(Request req, FeedSource feedSource) {
```
evansiroky, line 121: Please add JavaDoc for this method.

src/test/java/com/conveyal/datatools/manager/jobs/DeployJobTest.java
```diff
@@ -30,6 +30,8 @@
  *
  * Note: these tests require credentials on the IBI AWS account, which is why the class is tagged with Ignore (so that it is
  * not run with the rest of the test suite)
+ *
+ * FIXME: Does this need to extend UnitTest?
```
evansiroky, line 34: Yes, every test class that is not part of the e2e tests should extend the `com.conveyal.datatools.UnitTest` class. The `com.conveyal.datatools.UnitTest` class has some logic that will skip the tests if only e2e tests should be ran.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,192 @@
...
+     */
+    @After
+    public void tearDownTest() {
+        // Clean up
+        if (bartVersion != null) bartVersion.delete();
+        if (newBartVersion != null) newBartVersion.delete();
+    }
+
+    @Test
+    public void canReplaceGtfsPlusFile() throws IOException {
```
evansiroky, line 96: Would be useful to have some JavaDoc explaining the process and goal of this test case.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,192 @@
...
+        bartVersion = createFeedVersionFromGtfsZip(bart, "bart_old.zip");
+        // Fetch snapshot where modifications were made and create new version from it.
+        Snapshot snapshotWithModifications = bart.retrieveSnapshots().iterator().next();
+        CreateFeedVersionFromSnapshotJob newVersionJob = new CreateFeedVersionFromSnapshotJob(bart, snapshotWithModifications, user);
+        newVersionJob.run();
+        // Grab the modified version and check that the trips count matches expectation.
+        FeedVersion newVersion = bart.retrieveLatest();
+        assertEquals(
+            "trips count for transformed feed should be decreased by the # of records matched by the query",
+            bartVersion.feedLoadResult.trips.rowCount - 1041,
```
evansiroky, line 138: Using the BART GTFS for these tests is somewhat problematic here. For starters, the BART GTFS file contains 33k records in the stop_times.txt, so it will take longer to run than a GTFS feed with the minimal amount of data needed to demonstrate test case functionality. Next up, with so much data it is hard to verify stuff like this. Before making this magic number, did we look at all 2,512 trips in the feed to see if exactly 1,041 less should be present? Finally, it's not easy to view changes in version control to a zip file if it has to change.

I feel these tests would be much better to reason about if they all used a GTFS feed obtained from a new `fake-agency-*` folder that has close to the minimum amount of data required to demonstrate applicable test case functionality.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,192 @@
...
+            newVersion.feedLoadResult.trips.rowCount
+        );
+    }
+
+    @Test
+    public void replaceGtfsPlusFileFailsIfSourceIsMissing() {
```
evansiroky, line 144: Is there a list of all transformations and a functional spec of what each should do somewhere? I don't see it in the issue referenced from this PR (#544).

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
...
+        }
+
+        // FIXME: Should we overwrite the input GTFS dataset if transforming in place?
+
         // TODO: Any other activities that need to be run (e.g., module-specific activities).
+        if (DataManager.isModuleEnabled("deployment")) {
+//            Project project = feedSource.retrieveProject();
+//            if (project.autoDeployId)
+            // TODO: Get deployment, update feed version for feed source, and kick off deployment to server that
+            //  deployment is currently pointed at.
```
evansiroky, line 123: This seems like it needs more thought than just a TODO. In the case of a project with many feed sources, you might want to run something that updates all feed sources and then deploys as needed.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
...
+            // 2. We only have db transformations. These will always apply to a cloned snapshot of a feed version to
+            // avoid the primary db namespace having different contents from the zip file. Possibilities here are that
+            // the snapshot can end up hanging (i.e., not published) or once all of the transformations have been
+            // applied, we can create a new feed version from the snapshot (retrieval method being produced in house??).
+            // 3. Let's say we have zip transformations and db transformations.
+            // 3a: Zip transformations apply to imported version, db transformations apply to snapshot or get published (no big deal)
+            // 3b: Zip transformations apply to cloned version, db transformations apply to snapshot or get published.
+            //  ---- This case is a little trickier, because we could end up with two cloned_versions. Let's say we want
+            //  a SQL query to run on the original data, create a new GTFS file, and then replace a GTFS+ file from some
+            //  other version. FTR1 = clone
```
evansiroky, line 145: This whole block of comments confused me. A lot of it doesn't seem to necessarily apply to the immediate section of code below.

Also, what does `FTR1 = clone` mean?

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
...
         addNextJob(new LoadFeedJob(feedVersion, owner, isNewVersion));
 
         // Next, validate the feed.
         addNextJob(new ValidateFeedJob(feedVersion, owner, isNewVersion));
 
+        // We only need to snapshot the feed if there are transformations at the database level. In the case that there
+        // are, the snapshot namespace will be the target of these modifications. If we were to apply the modifications
+        // to the original feed version's namespace, the namespace and the zip file would become out of sync, which
+        // violates a core expectation for this application: that each zip file represents exactly its database
+        // representation.
```
evansiroky, line 98: This confuses me somewhat, because there is a possibility to transform a zip file in a previous block of code in this method with a ZipTransformation. Wouldn't this violate the stated expectation that each zip file that is inputted has a database representation?

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
     @Override
     public void jobLogic () {
         LOG.info("Processing feed for {}", feedVersion.id);
+        feedSource = feedVersion.parentFeedSource();
+        // If there are any feed transformations that apply to the VERSION_CLONE retrieval method, we need to add a
+        // stage to jobFinished (perhaps) to clone the feed version (as long as this feed version has not already been
+        // cloned). Once that feed version is processed, the appropriate transformations will apply themselves.
+        shouldClone = !feedVersion.retrievalMethod.equals(VERSION_CLONE) &&
+            feedSource.hasRulesForRetrievalMethod(VERSION_CLONE);
```
evansiroky, line 73: Are there any Feed Transformations that apply to version clones? I'm not sure if I found any. Therefore, it's kinda hard to see the purpose of this.

br648:
@landonreed Unable to run due to DB auth issues even though username and password are provided.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -113,6 +118,23 @@ private static FeedSource createFeedSource(Request req, Response res) throws IOE
...
+        if (feedSource.name == null || feedSource.name.isEmpty()) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Feed source name must be provided");
+            return false;
+        }
+        List<FeedSource.FeedRetrievalMethod> retrievalMethods = feedSource.transformRules.stream()
+            .map(rule -> rule.retrievalMethods)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
+        Set<FeedSource.FeedRetrievalMethod> retrievalMethodSet = new HashSet<>(retrievalMethods);
+        if (retrievalMethods.size() > retrievalMethodSet.size()) {
```
binh-dam-ibigroup, line 131: Could these two lines be merged into this?
```java
.collect(Collectors.toSet());
```

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -163,49 +124,63 @@ private static String getS3Key (String id) {
         return s3Prefix + id;
     }
 
-    public String getPathToFeed (String id) {
-        return new File(path, id).getAbsolutePath();
+    /**
+     * Get the File for the provided feed version ID (or other entity ID, depending on the feed store's context).
+     */
+    public File getPathToFeed(String id) {
```
binh-dam-ibigroup, line 130: Rename to `getFeedFile(String id)`.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
     @Override
     public void jobLogic () {
         LOG.info("Processing feed for {}", feedVersion.id);
+        feedSource = feedVersion.parentFeedSource();
```
binh-dam-ibigroup, line 68: Field `feedSource` should be initialized in the constructor.

binh-dam-ibigroup: Also, with `feedSource` initialized in the constructor, method `getFeedSourceId()` can also use it.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -48,21 +65,102 @@ public String getFeedSourceId () {
...
+        // If there are any feed transformations that apply to the VERSION_CLONE retrieval method, we need to add a
+        // stage to jobFinished (perhaps) to clone the feed version (as long as this feed version has not already been
+        // cloned). Once that feed version is processed, the appropriate transformations will apply themselves.
+        shouldClone = !feedVersion.retrievalMethod.equals(VERSION_CLONE) &&
```
binh-dam-ibigroup, line 72: Field `shouldClone` should be initialized in the constructor and made `final`.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -536,11 +534,20 @@ public void renumberSnapshots() {
      * post-processing a single version).
      */
     public enum FeedRetrievalMethod {
```
binh-dam-ibigroup, line 536: This might be large enough now to be in its own file.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -580,4 +587,35 @@ public void delete() {
...
+        return getRulesForRetrievalMethod(retrievalMethod) != null;
+    }
+
+    public FeedTransformRules getRulesForRetrievalMethod(FeedRetrievalMethod retrievalMethod) {
+        FeedTransformRules ruleSet = null;
+        for (FeedTransformRules rule : transformRules) {
+            if (rule.hasRetrievalMethod(retrievalMethod)) {
+                if (ruleSet != null) {
+                    LOG.warn("Found multiple transform rule sets for retrieval method {} (feed source {}). Defaulting to first found.", retrievalMethod, this.name);
+                    continue;
```
binh-dam-ibigroup, line 605: break instead?

binh-dam-ibigroup: Another way to implement the search would be with stream.filter()

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -235,8 +258,8 @@ public void load(MonitorableJob.Status status, boolean isNewVersion) {
             if (gtfsFile.length() == 0) {
                 throw new IOException("Empty GTFS file supplied");
             }
-            // If feed version has not been hashed, hash it here.
-            if (hash == null) hash = HashUtils.hashFile(gtfsFile);
+            // If somehow feed version has not been taken characteristics from GTFS file, handle this here.
```
binh-dam-ibigroup, line 261: This comment should be clarified.

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -424,4 +426,31 @@ public void delete() {
...
+    /**
+     * Assign characteristics from a new GTFS file to the feed version. This should generally be called directly after
+     * constructing the feed version (assuming the GTFS file is available). Characteristics set from file include:
+     * - last modified (file timestamp)
+     * - length (file size)
+     * - hash
+     * @param newGtfsFile   the new GTFS file
+     * @param lastModifiedOverride  optional override of the file's last modified value
+     */
+    public void assignGtfsFile(File newGtfsFile, Long lastModifiedOverride) {
```
binh-dam-ibigroup, line 439: Suggestion: rename to `assignGtfsFileAttributes` for clarity.

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -424,4 +426,31 @@ public void delete() {
...
+            fileTimestamp = newGtfsFile.lastModified();
+        }
+        fileSize = newGtfsFile.length();
+        if (hash == null) hash = HashUtils.hashFile(newGtfsFile);
+    }
+
+    /**
+     * Convenience wrapper for {@link #assignGtfsFile} that does not override file's last modified.
+     */
+    public void assignGtfsFile(File newGtfsFile) {
```
binh-dam-ibigroup, line 453: Rename this one per my above comment too.

src/main/java/com/conveyal/datatools/manager/models/TransformType.java
```diff
@@ -0,0 +1,10 @@
+package com.conveyal.datatools.manager.models;
+
+public enum TransformType {
+    REPLACE_FILE_FROM_STRING, // NYSDOT case for feed_info.txt
```
binh-dam-ibigroup, line 4: Use JavaDoc comment style.

src/main/java/com/conveyal/datatools/manager/models/transform/FeedTransformRules.java
```diff
@@ -0,0 +1,74 @@
...
+     * snapshot).
+     */
+    public boolean createNewVersion;
+
+    /** no-arg constructor for serialization */
+    public FeedTransformRules() {}
+
+    /** Default constructor to create rule set with the passed in transformations */
+    public FeedTransformRules(FeedTransformation ...transformations) {
+        // Default retrieval method to apply to is fetch/manual upload. In other words, transformations attached to this
```
binh-dam-ibigroup, line 37: Methods is plural here.

src/main/java/com/conveyal/datatools/manager/models/transform/FeedTransformTarget.java
```diff
@@ -0,0 +1,32 @@
...
+    public String snapshotId;
+
+    public FeedTransformTarget(File gtfsFile) {
+        this.gtfsFile = gtfsFile;
+    }
+
+    public FeedTransformTarget(String targetSnapshotId) {
+        this.snapshotId = targetSnapshotId;
+    }
+
```
binh-dam-ibigroup, line 24: This might need to be split into sub classes, such as `FeedTransformFileTarget` and `FeedTransformDbTarget`, to parallel the Zip and Db transforms.

binh-dam-ibigroup: Also, db operations are not valid with zip targets and vice-versa.

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromStringTransformation.java
```diff
@@ -0,0 +1,57 @@
...
+    public static ReplaceFileFromStringTransformation create(String csvData, String table) {
+        ReplaceFileFromStringTransformation transformation = new ReplaceFileFromStringTransformation();
+        transformation.csvData = csvData;
+        transformation.table = table;
+        return transformation;
+    }
+
+    @Override
+    public void transform(FeedTransformTarget target, MonitorableJob.Status status) {
+        // TODO: Refactor into validation code?
```
binh-dam-ibigroup, line 29: As I mentioned in `FeedTransformTarget`, you could only accept the zip version of it, with the checks embedded within it.

src/main/java/com/conveyal/datatools/manager/models/transform/FeedTransformTarget.java
binh-dam-ibigroup: After reading implementations, I am going to request the above split.

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -16,25 +16,24 @@
...
-import java.util.ArrayList;
-import java.util.List;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 import static com.conveyal.datatools.manager.DataManager.hasConfigProperty;
 
 /**
- * Store a feed on the file system or s3
+ * Store a feed on the file system or s3.
```
binh-dam-ibigroup, line 36: S3 capital S.

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -163,49 +124,63 @@ private static String getS3Key (String id) {
...
-            // FIXME: Figure out how to manage temp files created here. Currently, deleteOnExit is called in createTempFile
             try {
-                return createTempFile(id, objectData);
+                Path path = Files.createTempFile(id, null);
+                // FIXME: Figure out how to manage temp files created here. Currently, we just call deleteOnExit, but
+                //  this will only delete the file once the java process stops.
+                File tempFile = path.toFile();
+                tempFile.deleteOnExit();
+                ByteStreams.copy(objectData, new FileOutputStream(tempFile));
+                return tempFile;
```
binh-dam-ibigroup, line 163: Keep this inside `createTempFile`

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
binh-dam-ibigroup: This test should be in its own file.

binh-dam-ibigroup: I think there should be a file for all the transformations tested.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
binh-dam-ibigroup: Not a definite answer, but I'd think cloning should be handled by the owner object of the job, so that jobs don't have to deal with retrievalMethod.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
landonreed: No, I don't think so. I want to explicitly check that the list is not larger than the set (i.e., that there are no duplicates). I'll add a comment for clarity.

src/main/java/com/conveyal/datatools/manager/models/transform/FeedTransformTarget.java
landonreed: Great suggestion, @binh-dam-ibigroup.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
landonreed: @evansiroky, I just updated the issue https://github.com/ibi-group/datatools-ui/issues/544 with the transformation requirements. The `DeleteRecordsTransformation` isn't a requirement, but more of a POC for us to further develop transformations applied to db snapshots. It may be replaced in time with something else (but there is no UI element designed for it intentionally).

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
landonreed: What would you like to see instead of the TODO? A GitHub issue perhaps? This hasn't been built, but was intended just as a placeholder for an idea I had.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
landonreed: Sorry this code block was for me when I was figuring out how to structure everything. I will move to javadoc for jobLogic and clarify.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
landonreed: In the case that the zip transformation applies to incoming GTFS. That incoming GTFS is modified before it is ever loaded into the database, so the resulting zip and db match one another. Your comment does make me think of a potential pitfall to this approach: If we're fetching the feed automatically and applying a zip transformation, we may not be able to recognize the next time we fetch GTFS that the version we're loading is the same version we previously loaded (because we have modified it).

landonreed:
@evansiroky and @binh-dam-ibigroup, I think I responded to all of your (great) feedback. Let me know if you want more clarity anywhere. I did add a VERSION_CLONE test to hopefully clarify the intent there (but maybe there's a better way). Also, one pending issue is the need to test whether GTFS zips that are transformed do not interfere with future feed fetches. Hoping to address that soon...

@br648, I'm not too sure how to replicate your issue. Seems like it might just be a windows vs. *nix OS issue?

pom.xml
```diff
@@ -250,7 +250,7 @@
         <dependency>
             <groupId>com.conveyal</groupId>
             <artifactId>gtfs-lib</artifactId>
-            <version>6.0.1</version>
+            <version>6.0.3</version>
```
evansiroky, line 253: This is why I don't like MTC branches diverging from dev a bunch. We end up reviewing code we've already reviewed.

src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
```diff
@@ -494,22 +492,29 @@ private static String deploy (Request req, Response res) {
...
-        put(apiPrefix + "secure/deployments/:id", DeploymentController::updateDeployment, json::write);
-        post(apiPrefix + "secure/deployments/fromfeedsource/:id", DeploymentController::createDeploymentFromFeedSource, json::write);
+        get(apiPrefix + "secure/deployments/:id/ec2", DeploymentController::fetchEC2InstanceSummaries, slimJson::write);
+        delete(apiPrefix + "secure/deployments/:id/ec2", DeploymentController::terminateEC2InstanceForDeployment, slimJson::write);
+        get(apiPrefix + "secure/deployments/:id", DeploymentController::getDeployment, fullJson::write);
+        delete(apiPrefix + "secure/deployments/:id", DeploymentController::deleteDeployment, fullJson::write);
+        get(apiPrefix + "secure/deployments", DeploymentController::getAllDeployments, slimJson::write);
+        post(apiPrefix + "secure/deployments", DeploymentController::createDeployment, fullJson::write);
+        put(apiPrefix + "secure/deployments/:id", DeploymentController::updateDeployment, fullJson::write);
+        post(apiPrefix + "secure/deployments/fromfeedsource/:id", DeploymentController::createDeploymentFromFeedSource, fullJson::write);
```
evansiroky, line 518: It is nearly impossible to tell what this PR is about anymore since it contains a whole bunch of stuff from the dev branch.

src/main/java/com/conveyal/datatools/manager/models/transform/DeleteRecordsTransformation.java
```diff
@@ -0,0 +1,72 @@
...
+            status.fail(String.format("Cannot find snapshot to transform (id=%s)", dbTarget.snapshotId));
+            return;
+        }
+        // Create a new SQL connection, construct the SQL statement, execute and commit results.
+        try (Connection connection = GTFS_DATA_SOURCE.getConnection()) {
+            String deleteSql = String.format(
+                "delete from %s.%s where %s in (%s)",
+                snapshot.namespace,
+                table,
+                matchField,
```
evansiroky, line 56: Both `table` and `matchField` seem like possibilities for SQL injection. As far as I can tell, it looks like they could come from direct user input. Therefore, before making this query, we need to verify if `table` is a valid table and if `matchField` is a valid field in the requested table.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,271 @@
...
+            feedSource,
+            zipFolderFiles("fake-agency-with-only-calendar")
+        );
+        // Replace file transformation runs before feed is loaded into database.
+        // Note: stop_attributes.txt is a GTFS+ file found in BART's feed.
+        FeedTransformation transformation = ReplaceFileFromVersionTransformation.create(sourceVersion.id, "realtime_routes");
+        FeedTransformRules transformRules = new FeedTransformRules(transformation);
+        feedSource.transformRules.add(transformRules);
+        Persistence.feedSources.replace(feedSource.id, feedSource);
+        // Create new BART version (note: bart_new.zip GTFS file has been stripped of stop_attributes.txt)
```
evansiroky, line 219: Outdated reference to BART here.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,271 @@
...
+        final String feedInfoContent = generateFeedInfo(feedId);
+        sourceVersion = createFeedVersion(
+            feedSource,
+            zipFolderFiles("fake-agency-with-only-calendar")
+        );
+        FeedTransformation transformation = ReplaceFileFromStringTransformation.create(feedInfoContent, "feed_info");
+        FeedTransformRules transformRules = new FeedTransformRules(transformation);
+        feedSource.transformRules.add(transformRules);
+        Persistence.feedSources.replace(feedSource.id, feedSource);
+        // Create new BART version (note: bart_new.zip GTFS file has been stripped of
```
evansiroky, line 241: Outdated reference to BART here.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -42,20 +66,110 @@ public String getFeedVersionId () {
...
+            }
+        }
+
+        if (shouldClone) {
+            // If it was determined that the final version should be cloned (because of feed transformations that
+            // must operate on a cloned version), handle that here. The cloned version will then apply the
+            // transformations in its process feed job.
+            try {
+                // Create a new version for the clone.
+                FeedVersion newFeedVersion = new FeedVersion(feedSource, VERSION_CLONE);
```
evansiroky, line 143: So how many FeedVersions will get created here? This seems like it would have the potential to clutter things up. Let's say a few transformations exist on a feed that is fetched nightly. Instead of just one single FeedVersion being created, there could be multiple... along with additional snapshots. Seems like we may want to just drop this expectation that a FeedVersion exactly represents an incoming GTFS file if there are transforms that get applied. We should still save the original GTFS and perhaps the pre-processed input GTFS, but having a bunch of FeedVersions and intermediate snapshots created seems like we're trying to use existing tools instead of making a proper new tool.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    @Test
+    public void canCloneZipFileAndTransform() throws IOException, SQLException {
+        // Generate random UUID for feedId, which gets placed into the csv data.
+        final String feedId = UUID.randomUUID().toString();
+        final String feedInfoContent = generateFeedInfo(feedId);
+        FeedTransformation transformation = ReplaceFileFromStringTransformation.create(feedInfoContent, "feed_info");
+        // Create transform rules for VERSION_CLONE retrieval method.
+        FeedTransformRules transformRules = new FeedTransformRules(transformation, FeedRetrievalMethod.VERSION_CLONE);
```
evansiroky, line 169: What real-life use case is this trying to model? Why would someone want to do this instead of the transformation done in the `canReplaceFeedInfo` test case?

src/main/java/com/conveyal/datatools/manager/models/FeedRetrievalMethod.java
```diff
@@ -0,0 +1,25 @@
...
+    MANUALLY_UPLOADED,
+    /** Feed produced in-house in a GTFS Editor instance (i.e., from a database snapshot). */
+    PRODUCED_IN_HOUSE,
+    /** Feed produced in-house in the GTFS+ Editor. */
+    PRODUCED_IN_HOUSE_GTFS_PLUS,
+    /** Feed produced from a regional merge (e.g., merging all feeds from a project). */
+    REGIONAL_MERGE,
+    /** Feed produced by merging two versions that span different service periods. */
+    SERVICE_PERIOD_MERGE,
+    /** Feed produced by cloning an existing version. */
```
evansiroky, line 23: Still confused by how exactly a VERSION_CLONE version gets created. It appears that it can be from either a Snapshot or post-processing a fetched or uploaded version. I feel like this comment can be expanded upon a bit.

src/main/java/com/conveyal/datatools/manager/models/transform/ZipTransformation.java
```diff
@@ -0,0 +1,12 @@
+package com.conveyal.datatools.manager.models.transform;
+
+/**
+ * This is an abstract class that represents a transformation that should apply to a GTFS in zip form. In other
+ * words, subclasses will provide a transform override method that acts directly on the zip file. Sample fields
+ * csvData and sourceVersionId can be used to reference replacement string or file that should be passed to the target
+ * zip file of the transformation.
+ */
+public abstract class ZipTransformation extends FeedTransformation {
```
evansiroky, line 9: Not sure how much value this file adds. Especially since the `csvData` field isn't even used by the `ReplaceFileFromVersionTransformation` transformer. Another option might be to have a `FeedTransformation` have an enumeration of the stage of when the transformation is supposed to take place.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
```diff
@@ -42,20 +66,110 @@ public String getFeedVersionId () {
...
+     * 2. We only have db transformations. These will always apply to a cloned snapshot of a feed version to
+     *    prevent the feed version's namespace from having different contents from the zip file. Following this, the
+     *    snapshot can either end up hanging (i.e., not published) or once all of the transformations have been
+     *    applied, we can create a new feed version from the snapshot (retrieval method being produced in house??).
+     * 3. Let's say we have zip transformations and db transformations.
+     *    a. Zip transformations apply to imported version, db transformations apply to snapshot or get published (no big deal)
+     *    b. Zip transformations apply to cloned version, db transformations apply to snapshot or get published.
+     *       This case is a little trickier, because we could end up with two cloned_versions. Let's say we want
+     *       a SQL query to run on the original data, create a new GTFS file, and then replace a GTFS+ file from some
+     *       other version. FTR1 = clone
```
evansiroky, line 90: Still don't know what `FTR1 = clone` means.

It would probably be much better to visualize this with a flow chart. Either an ASCII comment indicating at what stage certain transformations might occur or a link to docs that has a diagram of boxes and lines indicating all this.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
evansiroky: Yeah, I'd remove this comment and make an issue. Automated deployments isn't even a part of the MTC wishlist in https://github.com/ibi-group/datatools-ui/issues/544.

br648:
@landonreed - This is probably Windows related, but I'm unable to run any tests because of file encoding. I've set this in the run VM options (-Dfile.encoding=UTF-8) and can clearly see this as part of the Java parameters in the logging, but I'm not convinced this is the correct place to set it. If I remove it from the VM options it is still present in the logging. Perhaps the file encoding needs to be  set under Maven when testing, but I'm not sure where?

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
landonreed: I don't know it means either. Sorry, I meant to remove that mysterious comment.

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromStringTransformation.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void transform(FeedTransformTarget target, MonitorableJob.Status status) {
+        if (!(target instanceof FeedTransformZipTarget)) {
+            status.fail("Target must be FeedTransformZipTarget.");
+            return;
+        }
+        // Cast transform target to zip flavor.
+        FeedTransformZipTarget zipTarget = (FeedTransformZipTarget)target;
```
binh-dam-ibigroup, line 34: Same remark for making `target` be of type `FeedTransformZipTarget`.

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromVersionTransformation.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public void transform(FeedTransformTarget target, MonitorableJob.Status status) {
+        if (!(target instanceof FeedTransformZipTarget)) {
+            status.fail("Target must be FeedTransformZipTarget.");
+            return;
+        }
+        // Cast transform target to zip flavor.
+        FeedTransformZipTarget zipTarget = (FeedTransformZipTarget)target;
```
binh-dam-ibigroup, line 40: Same remark.

src/main/java/com/conveyal/datatools/manager/jobs/ProcessSingleFeedJob.java
landonreed: Instead of creating all of these feed versions and snapshots, after discussion we will just try to create one to avoid confusion for users. We can just show a transformation summary for any version that was modified.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
landonreed: Will address in modifications to how versions are created by transformations.

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromStringTransformation.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        if (table == null) {
+            status.fail("Must specify transformation table name.");
+            return;
+        }
+        String tableName = table + ".txt";
+        String tableNamePath = "/" + tableName;
+        // Run the replace transformation
+        Path targetZipPath = Paths.get(zipTarget.gtfsFile.getAbsolutePath());
+        try( FileSystem targetZipFs = FileSystems.newFileSystem(targetZipPath, null) ){
```
br648, line 48: Space between try( and ){

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromVersionTransformation.java
```diff
@@ -0,0 +1,72 @@
...
+        String tableNamePath = "/" + tableName;
+
+        // Run the replace transformation
+        Path sourceZipPath = Paths.get(sourceVersion.retrieveGtfsFile().getAbsolutePath());
+        try (FileSystem sourceZipFs = FileSystems.newFileSystem(sourceZipPath, null)) {
+            // If the source txt file does not exist, NoSuchFileException will be thrown and caught below.
+            Path sourceTxtFilePath = sourceZipFs.getPath(tableNamePath);
+            Path targetZipPath = Paths.get(zipTarget.gtfsFile.getAbsolutePath());
+            LOG.info("Replacing file {} in zip file {} with source {}", tableNamePath, targetZipPath.getFileName(), sourceVersion.id);
+            try( FileSystem targetZipFs = FileSystems.newFileSystem(targetZipPath, null) ){
```
br648, line 60: Space between try( and ){

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -100,53 +98,16 @@ private static File getPath (String pathString) {
...
-            return s3Client.doesObjectExist(s3Bucket, getS3Key(id)) ? s3Client.getObjectMetadata(s3Bucket, getS3Key(id)).getContentLength() : null;
-        }
-        else {
-            File feed = getFeed(id);
-            return feed != null ? feed.length() : null;
+        // Always delete local copy (whether storing exclusively on local disk or using s3).
+        File feed = getLocalFeed(id);
+        if (feed != null) {
+            boolean deleted = feed.delete();
+            if (!deleted) LOG.warn("GTFS file {} not deleted. This may contribute to storage space shortages.", feed.getAbsolutePath());
```
br648, line 110: Brackets.

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -163,49 +124,58 @@ private static String getS3Key (String id) {
...
-        File feed = new File(path, id);
-        // Whether storing locally or on s3, return local version if it exists.
-        // Also, don't let folks retrieveById feeds outside of the directory
-        if (feed.getParentFile().equals(path) && feed.exists()) {
-            return feed;
-        }
-
+        // Whether storing locally or on s3, first try returning the local copy if it exists.
+        File feed = getLocalFeed(id);
+        if (feed != null) return feed;
```
br648, line 140: Brackets.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,271 @@
...
+        if (sourceVersion != null) sourceVersion.delete();
+        if (targetVersion != null) targetVersion.delete();
+    }
+
+    /**
+     * Test that a {@link ReplaceFileFromVersionTransformation} will successfully add a GTFS+ file found in the source version
+     * into the target version's GTFS file.
+     */
+    @Test
+    public void canReplaceGtfsPlusFileFromVersion() throws IOException {
```
br648, line 104: On Windows you get the following warning which in turn fails this test:

java.nio.file.FileSystemException: C:\tmp\Test_Feed_Tue_Jun_30_145808_BST_2020-20200630T145809-01-619a856f-4b4d-4abb-a267-8eb353141241-1af5e4ae-c948-4320-8843-85c6beeaf5de.zip: The process cannot access the file because it is being used by another process.

This is for reference. Perhaps if I move on to working on datatools I'll attempt to resolve.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,271 @@
...
+        targetVersion = createFeedVersion(
+            feedSource,
+            zipFolderFiles("fake-agency-with-only-calendar-dates")
+        );
+        // TODO Check that new version has stop_attributes file that matches source version's copy.
+        assertThat(targetVersion.validationResult, Matchers.nullValue());
+    }
+
+    @Test
+    public void canReplaceFeedInfo() throws SQLException, IOException {
```
br648, line 229: On Windows you get the following warning which in turn fails this test:

java.nio.file.FileSystemException: C:\tmp\Test_Feed_Tue_Jun_30_145808_BST_2020-20200630T145809-01-619a856f-4b4d-4abb-a267-8eb353141241-1af5e4ae-c948-4320-8843-85c6beeaf5de.zip: The process cannot access the file because it is being used by another process.

This is for reference. Perhaps if I move on to working on datatools I'll attempt to resolve.

src/main/java/com/conveyal/datatools/manager/models/transform/DeleteRecordsTransformation.java
```diff
@@ -0,0 +1,72 @@
...
+        transformation.matchValues = matchValues;
+        return transformation;
+    }
+
+    @Override
+    public void transform(FeedTransformTarget target, MonitorableJob.Status status) {
+        if (!(target instanceof FeedTransformDbTarget)) {
+            status.fail("Target must be FeedTransformDbTarget.");
+            return;
+        }
```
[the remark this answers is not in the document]
landonreed: I don't think the polymorphism will allow this because we need the `FeedTransformTarget` param in the transform method signature to match the super class

src/main/java/com/conveyal/datatools/manager/models/transform/ReplaceFileFromStringTransformation.java
landonreed: See previous response.

src/main/java/com/conveyal/datatools/common/status/MonitorableJob.java
```diff
@@ -176,12 +177,13 @@ public void run () {
                 // because the error presumably already occurred and has a better error message.
                 cancel(cancelMessage);
             }
-            // Run final steps of job pending completion or error. Note: any tasks that depend on job success should
-            // check job status to determine if final step should be executed (e.g., storing feed version in MongoDB).
-            // TODO: should we add separate hooks depending on state of job/sub-tasks (e.g., success, catch, finally)
             // Complete the job (as success if no errors encountered, as failure otherwise).
             if (!parentJobErrored && !subTaskErrored) status.completeSuccessfully("Job complete!");
             else status.complete(true);
```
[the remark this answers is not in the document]
landonreed: The style guide we follow says it's ok for brackets to be missing if the entire block is all on a single line.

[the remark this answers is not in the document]
landonreed: Actually, I'm not really sure if that's in writing anywhere, but it's a convention we've adopted for some reason.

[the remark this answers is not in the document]
br648: ok no worries. Will bear that in mind with future reviews and coding. Still molding myself to OTP dev style!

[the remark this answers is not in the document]
binh-dam-ibigroup: If it is all on one line then there are no ambiguities.

src/main/java/com/conveyal/datatools/manager/models/transform/DeleteRecordsTransformation.java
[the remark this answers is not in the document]
binh-dam-ibigroup: Could this work?
```java
class FeedTransformation<T extends FeedTransformTarget> ...
    public abstract void transform(T target, ...)
```
followed by
```java
public abstract class DbTransformation extends FeedTransformation<FeedTransformDbTarget> {
...
public class DeleteRecordsTransformation extends DbTransformation {
    ...
    @Override
    public void transform(FeedTransformDbTarget target, ...) {

```

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -0,0 +1,270 @@
...
+import java.util.zip.ZipFile;
+
+import static com.conveyal.datatools.TestUtils.appendDate;
+import static com.conveyal.datatools.TestUtils.assertThatSqlCountQueryYieldsExpectedCount;
+import static com.conveyal.datatools.TestUtils.createFeedVersion;
+import static com.conveyal.datatools.TestUtils.zipFolderFiles;
+import static com.conveyal.datatools.manager.models.FeedRetrievalMethod.MANUALLY_UPLOADED;
+import static com.conveyal.datatools.manager.models.FeedRetrievalMethod.VERSION_CLONE;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
```
binh-dam-ibigroup, line 41: This import appears unused.

src/main/java/com/conveyal/datatools/manager/models/FeedTransformResult.java
```diff
@@ -0,0 +1,27 @@
...
+     * cannot be easily (de)serialized by MongoDB.
+     */
+    public List<TableTransformResult> tableTransformResults = new ArrayList<>();
+
+    public FeedTransformResult() {}
+
+    /**
+     * Shorthand to add a table transform result to the appropriate string-keyed table name.
+     */
+    public void addResultForTable (TableTransformResult tableTransformResult) {
```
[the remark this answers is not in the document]
landonreed: Perhaps not. It was useful during development when things kept changing on me, but it doesn't provide much of a shorthand, does it?

src/main/java/com/conveyal/datatools/manager/models/transform/FeedTransformTarget.java
```diff
@@ -1,13 +1,15 @@
...
 import com.conveyal.datatools.common.status.MonitorableJob;
+import com.conveyal.datatools.manager.models.FeedTransformResult;
 
 /**
  * Interface for the different targets that a {@link FeedTransformation} can apply to:
  * - zip file {@link FeedTransformZipTarget} and
  * - db snapshot {@link FeedTransformDbTarget} (contains snapshotId which is a proxy for the database namespace)
  */
 public interface FeedTransformTarget {
+    FeedTransformResult feedTransformResult = new FeedTransformResult();
```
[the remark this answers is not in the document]
landonreed: Yes, I think you're right. Good catch.

PR #330, opened 2020-07-24
Create job for recreating graph build image

evansiroky:
also parallelize ec2 validation checks

### Checklist

- [ ] The description lists all applicable issues this PR seeks to resolve

### Description

This PR does the following:

- refactors the validation of ec2 config to be parallelized and accessible from other classes
- adds a validation check to make sure the graph build instance name that should be recreated has a unique name
- adds a validation check to AMI IDs to make sure they have an `available` state
- adds validation of ec2 config to the beginning of a deploy job
- adds a new job of `RecreateBuildImageJob` that happens asynchronously from the rest of the deploy job

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
@@ -54,11 +54,17 @@ import spark.Response;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
 import static com.conveyal.datatools.common.utils.SparkUtils.getPOJOFromRequestBody;
@@ -298,17 +304,14 @@ public class ServerController {
                         s3Client = AWSUtils.getS3ClientForCredentials(getAWSCreds(), server.ec2Info.region);
                     }
                 }
-                validateInstanceType(server.ec2Info.instanceType, req);
-                validateInstanceType(server.ec2Info.buildInstanceType, req);
-                // Validate target group and get load balancer to validate subnetId and security group ID.
-                LoadBalancer loadBalancer = validateTargetGroupAndGetLoadBalancer(server.ec2Info, req, credentials);
-                validateSubnetId(loadBalancer, server.ec2Info, req, ec2Client);
-                validateSecurityGroupId(loadBalancer, server.ec2Info, req);
-                // Validate remaining AWS values.
-                validateIamInstanceProfileArn(server.ec2Info.iamInstanceProfileArn, req, iamClient);
-                validateKeyName(server.ec2Info.keyName, req, ec2Client);
-                validateAmiId(server.ec2Info.amiId, req, ec2Client);
-                validateAmiId(server.ec2Info.buildAmiId, req, ec2Client);
+                try {
+                    EC2ValidationResult result = validateEC2Config(server, credentials, ec2Client, iamClient);
+                    if (!result.isValid()) {
+                        logMessageAndHalt(req, 400, result.getMessage(), result.getException());
+                    }
+                } catch (ExecutionException | InterruptedException e) {
+                    logMessageAndHalt(req, 500, "Failed to validate EC2 config", e);
+                }
                 if (server.ec2Info.instanceCount < 0) server.ec2Info.instanceCount = 0;
             }
             // Server must have name.
@@ -326,6 +329,53 @@ public class ServerController {
             if (e instanceof HaltException) throw e;
             else logMessageAndHalt(req, 400, "Error encountered while validating server field", e);
         }
+    }
+
+    /**
+     * Asynchrnously validates all ec2 config of a particular OtpServer instance.
+     */
+    public static EC2ValidationResult validateEC2Config(
+        OtpServer server,
+        AWSStaticCredentialsProvider credentials,
+        AmazonEC2 ec2Client,
+        AmazonIdentityManagement iamClient
+    ) throws ExecutionException, InterruptedException {
+        List<Callable<EC2ValidationResult>> validationTasks = new ArrayList<>();
+        validationTasks.add(() -> validateInstanceType(server.ec2Info.instanceType));
+        validationTasks.add(() -> validateInstanceType(server.ec2Info.buildInstanceType));
+        validationTasks.add(() -> validateIamInstanceProfileArn(server.ec2Info.iamInstanceProfileArn, iamClient));
+        validationTasks.add(() -> validateKeyName(server.ec2Info.keyName, ec2Client));
+        validationTasks.add(() -> validateAmiId(server.ec2Info.amiId, ec2Client));
+        validationTasks.add(() -> validateAmiId(server.ec2Info.buildAmiId, ec2Client));
+        validationTasks.add(() -> validateGraphBuildReplacementAmiName(server.ec2Info, ec2Client));
+        // add the load balancer task to the end since it can produce aggregate messages
+        validationTasks.add(() -> validateTargetGrouptLoadBalancerSubnetIdAndSecurityGroup(
+            server.ec2Info,
+            credentials,
+            ec2Client
+        ));
+
+        // execute all tasks asynchronously in unique threads
+        ExecutorService pool = Executors.newFixedThreadPool(validationTasks.size());
+
+        // compile overall result
+        EC2ValidationResult result = new EC2ValidationResult();
+        // check each individual task result
+        for (Future<EC2ValidationResult> ec2ValidationResultFuture : pool.invokeAll(validationTasks)) {
+            EC2ValidationResult taskValidationResult = ec2ValidationResultFuture.get();
+            // check if task yielded a valid result
+            if (!taskValidationResult.isValid()) {
+                // task had an invalid result, check if overall validation result has been changed to false yet
+                if (result.isValid()) {
+                    // first invalid result. Write a header message.
+                    result.setInvalid("Invalid EC2 config for the following reasons!\n", new Exception());
+                }
+                // add to list of messages and exceptions
+                result.appendResult(taskValidationResult);
+            }
+        }
+        pool.shutdown();
+        return result;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -420,27 +524,34 @@ private static void validateSecurityGroupId(LoadBalancer loadBalancer, EC2Info e
             if (securityGroupId != null) {
                 // Set security group to the first value found attached to ELB.
                 ec2Info.securityGroupId = securityGroupId;
```
landonreed, line 526: Since we're using this in more contexts, is this attempt to set the security group still ok? Just checking.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -522,17 +639,44 @@ private static LoadBalancer getLoadBalancerForTargetGroup (EC2Info ec2Info, AWSS
      * Validate that ELB target group exists and is not empty and return associated load balancer for validating related
      * fields.
      */
-    private static LoadBalancer validateTargetGroupAndGetLoadBalancer(EC2Info ec2Info, Request req, AWSStaticCredentialsProvider credentials) {
+    private static EC2ValidationResult validateTargetGrouptLoadBalancerSubnetIdAndSecurityGroup(
```
landonreed, line 642: typo: 

```
validateTargetGrouptLoadBalancerSubnetIdAndSecurityGroup
                   ^
```

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -522,17 +639,44 @@ private static LoadBalancer getLoadBalancerForTargetGroup (EC2Info ec2Info, AWSS
...
+        loadBalancerValidationTasks.add(() -> validateSecurityGroupId(loadBalancer, ec2Info));
+        ExecutorService pool = Executors.newFixedThreadPool(loadBalancerValidationTasks.size());
+        for (Future<EC2ValidationResult> ec2ValidationResultFuture : pool.invokeAll(loadBalancerValidationTasks)) {
+            EC2ValidationResult loadBalancerValidationTaskResult = ec2ValidationResultFuture.get();
+            // check if task yielded a valid result
+            if (!loadBalancerValidationTaskResult.isValid()) {
+                // task had an invalid result, check if overall validation result has been changed to false yet
+                if (result.isValid()) {
+                    // first invalid result. Write a header message.
+                    result.setInvalid("Invalid EC2 load balancer config for the following reasons:\n", new Exception());
```
landonreed, line 672: Empty exception? Is this intentional?

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -548,33 +569,21 @@ private void replaceEC2Servers() {
...
-                            .withImageId(otpServer.ec2Info.buildAmiId);
-                        ec2.deregisterImage(deregisterImageRequest);
-                    }
-                    status.update("Updating Server build AMI info", 45);
-                    // Update OTP Server info
-                    otpServer.ec2Info.buildAmiId = createImageResult.getImageId();
-                    otpServer.ec2Info.recreateBuildImage = false;
-                    Persistence.servers.replace(otpServer.id, otpServer);
+                    // Begin a new job to create a graph build image so that can happen asynchronously
+                    RecreateBuildImageJob recreateBuildImageJob = new RecreateBuildImageJob(
```
landonreed, line 573: This will only happen in a reasonable amount of time if there are enough threads available in the heavyExecutor. If not, it could get queued up for a very long time. I'm not sure we want to do this...

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -328,6 +331,53 @@ private static void validateFields(Request req, OtpServer server) throws HaltExc
...
+        EC2ValidationResult result = new EC2ValidationResult();
+        // check each individual task result
+        for (Future<EC2ValidationResult> ec2ValidationResultFuture : pool.invokeAll(validationTasks)) {
+            EC2ValidationResult taskValidationResult = ec2ValidationResultFuture.get();
+            // check if task yielded a valid result
+            if (!taskValidationResult.isValid()) {
+                // task had an invalid result, check if overall validation result has been changed to false yet
+                if (result.isValid()) {
+                    // first invalid result. Write a header message.
+                    result.setInvalid("Invalid EC2 config for the following reasons!\n", new Exception());
```
landonreed, line 371: Intentionally blank exception? If so, please explain why.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -522,17 +639,44 @@ private static LoadBalancer getLoadBalancerForTargetGroup (EC2Info ec2Info, AWSS
...
+            return result;
+        }
+
+        // asynchronously execute the two validation tasks that depend on the load balancer info
+        List<Callable<EC2ValidationResult>> loadBalancerValidationTasks = new ArrayList<>();
+        loadBalancerValidationTasks.add(() -> validateSubnetId(loadBalancer, ec2Info, ec2Client));
+        loadBalancerValidationTasks.add(() -> validateSecurityGroupId(loadBalancer, ec2Info));
+        ExecutorService pool = Executors.newFixedThreadPool(loadBalancerValidationTasks.size());
+        for (Future<EC2ValidationResult> ec2ValidationResultFuture : pool.invokeAll(loadBalancerValidationTasks)) {
+            EC2ValidationResult loadBalancerValidationTaskResult = ec2ValidationResultFuture.get();
```
landonreed, line 666: This code seems very similar to the above code in validateEC2Config. Can it be DRYed out?

src/main/java/com/conveyal/datatools/manager/jobs/RecreateBuildImageJob.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void jobLogic() throws Exception {
+        status.update("Creating build image", 5);
+        // Create a new image of this instance.
+        CreateImageRequest createImageRequest = new CreateImageRequest()
+            .withInstanceId(graphBuildingInstances.get(0).getInstanceId())
+            .withName(otpServer.ec2Info.buildImageName)
+            .withDescription(otpServer.ec2Info.buildImageDescription);
+        CreateImageResult createImageResult = ec2.createImage(createImageRequest);
+        // Wait for image creation to complete
```
binh-dam-ibigroup, line 52: Add a comment on how long this process could take.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -328,6 +331,53 @@ private static void validateFields(Request req, OtpServer server) throws HaltExc
         }
     }
 
+    /**
+     * Asynchrnously validates all ec2 config of a particular OtpServer instance.
```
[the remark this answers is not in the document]
evansiroky: In my experience, this has resulted in faster validation times because everything is now happening asynchronously. Previously, each validation check would run synchronously which meant that each validation check needed to wait for the previous one to finish and since there are a lot of calls to various AWS services this would take a while. The amount of time it took to setup a thread pool was much less than the time to make each validation check synchronously.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
```diff
@@ -366,33 +416,76 @@ private static void verifyS3WritePermissions(OtpServer server, Request req, Amaz
...
+        }
+        return false;
+    }
+
+    /**
+     * Validates whether the replacement graph build image name is unique. Although it is possible to have duplicate AMI
+     * names when copying images, they must be unique when creating images.
+     * See https://forums.aws.amazon.com/message.jspa?messageID=845159
+     */
+    private static EC2ValidationResult validateGraphBuildReplacementAmiName(
```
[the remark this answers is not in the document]
evansiroky: I suppose that's another way to do it, but I kind of like letting the user define the name and checking it this way to be sure it'll work.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
evansiroky: um, I guess it is. Haven't seen any failures because of this.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
evansiroky: Hmmm. This makes me think we ought to have a 3rd type of executor for long-lived, but low CPU tasks and then we would put all of these deploy job tasks that are basically only making HTTP requests.

src/main/java/com/conveyal/datatools/manager/controllers/api/ServerController.java
evansiroky: They can't really be returned since they depend on first obtaining the load balancer info.

evansiroky:
I just did a little refactor, but have some open ideas on how to address some other comments. See this one in particular: https://github.com/ibi-group/datatools-server/pull/330#discussion_r465246478.

wolli:
Herzlichen Dank an Euch,
habe mich 1mal abgemeldet.
LG

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
landonreed: I think a deploy job could still be high in memory requirements - e.g., if it's using build graph over wire. Why not just have this RecreateBuildImageJob run in the lightExecutor? Should the lightExecutor have more threads at its disposal? I'm just not aware of all of thread safety concerns if we increase the thread count (insert Egyptian cotton joke here).

evansiroky:
@landonreed , I went ahead and made the RecreateBuildImageJob use a new SingleThreadExecutor that is internal to the DeployJob. The DeployJob already starts its own thread pools anyways when running MonitorServerStatusJobs. Choosing the right number of threads to create is always a balancing act of trying to figure out how much CPU/Memory each thread will need to utilize, but as long as the threads are just sitting there waiting for stuff and then making a check-in HTTP request every once in a while, it shouldn't be much of an issue. In [this StackOverflow answer](https://stackoverflow.com/a/763584/269834), people have been saying they can get Java 8+ to handle over 400,000 threads.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -569,24 +578,31 @@ private void replaceEC2Servers() {
```
landonreed, line 578: (Couldn't ref this line, so adding code ref below):

Is it still true that the graph build only instance will self-terminate in the user data? Also, should this be the end of the job or should it wait for recreateGraphBuildImageJob?

https://github.com/ibi-group/datatools-server/blob/06138de0bfc3abd6afa86c74e597a2c8fe70c6b6/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java#L569-L574

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -655,6 +671,21 @@ private void replaceEC2Servers() {
...
+            // Wait for a recreate graph build job to complete if one was started
+            if (recreateBuildImageExecutor != null) {
+                status.update("Waiting for recreate graph building image job to complete", 95);
+                while (!recreateBuildImageJob.status.completed) {
+                    try {
+                        // wait 1 second
+                        Thread.sleep(1000);
+                    } catch (InterruptedException e) {
+                        recreateBuildImageExecutor.shutdown();
+                        status.fail("An error occurred while waiting for the graph build image to be recreated", e);
```
landonreed, line 685: Should an exception here constitute a failure of the entire deploy job? Also does `recreateBuildImageJob` need to be failed in this catch block?

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -569,24 +578,31 @@ private void replaceEC2Servers() {
 
                 // Check if a new image of the instance with the completed graph build should be created.
                 if (otpServer.ec2Info.recreateBuildImage) {
```
landonreed, line 580: (Couldn't ref this line, so adding code ref below):

Just stumbled across this uncommented line. Do you recall what this persistence operation is intended to do? If so, could you add a comment?

https://github.com/ibi-group/datatools-server/blob/06138de0bfc3abd6afa86c74e597a2c8fe70c6b6/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java#L577

src/main/java/com/conveyal/datatools/manager/jobs/RecreateBuildImageJob.java
```diff
@@ -63,14 +64,28 @@ public void jobLogic() throws Exception {
                     // See https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/model/ImageState.html
                     String imageState = image.getState().toLowerCase();
                     if (imageState.equals("pending")) {
+                        if (System.currentTimeMillis() - status.startTime > TimeUnit.HOURS.toMillis(1)) {
+                            terminateInstanceAndFailWithMessage(
+                                "It has taken over an hour for the graph build image to be created! Check the AWS console to see if the image was created successfully."
```
binh-dam-ibigroup, line 69: This log message is ambiguous. Has the graph completed at this stage?

binh-dam-ibigroup: Change the log message to: `The graph image build started over an hour ago. Check in the AWS console whether the image was created successfully.`

landonreed: `graph image build` is not quite accurate. It's recreating the `graph build image` (i.e., the AMI for the graph build instance).

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
evansiroky: It won't automatically shut down, I will update the comment.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
evansiroky: This doesn't seem to be needed anymore, so I removed it.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
evansiroky: Yes, I think it should. My logic is that if one part of the DeployJob fails (including this recreate build image job), then the entire job should be failed. There should be enough feedback via the other jobs to show the user that just the recreate build image job failed if that was the only failure.

I added some logic to also fail the build image job just in case.

Issue #337, opened 2020-09-21
Modify existing user metadata if Auth0 user already exists when creating user


landonreed:
## Observed behavior

When an Auth0 user already exists in the tenant when creating a new user in data tools, an error is encountered.

## Expected behavior

Instead of failing to create the new user, the existing Auth0 user's metadata should be modified to include the permissions for the new data tools user account.

## Steps to reproduce the problem

1. Create user through Auth0 portal for email account (with no user metadata).
2. Open data tools user admin, and attempt to create user for same email.
3. See error.

## Any special notes on configuration used

N/A

## Version of datatools-server and datatools-ui if applicable (exact commit hash or branch name)

latest dev

binh-dam-ibigroup:
Closing. This was fixed by #338.

PR #338, opened 2020-09-24
Update permissions for existing Auth0 users

br648:
### Checklist

- [x] All tests and CI builds passing ~(Failing, but is failing for other PRs)~

### Description

PR to cover this issue: https://github.com/ibi-group/datatools-server/issues/337

Note: I've only tested this via the datatools ui admin page which calls the method UserController -> createUser. There is another method in this class called createPublicUser, but I don't know how to trigger a call to it. I have updated the code in this method to handle existing Auth0 users, but I have not tested it. THE createPublicUser IS NO LONGER NEEDED AND HAS BEEN REMOVED.


binh-dam-ibigroup (changes requested):
Please address the CI test failures.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Users.java
```diff
@@ -45,16 +45,22 @@
...
+        return getUrl(searchQuery, page, perPage, includeTotals, true);
+    }
+
     /**
      * Constructs a user search query URL.
      * @param searchQuery   search query to perform (null value implies default query)
      * @param page          which page of users to return
      * @param perPage       number of users to return per page
      * @param includeTotals whether to include the total number of users in search results
+     * @param includeDefaultQuery whether to include the default query in the search criteria
```
landonreed, line 58: This param name might make more sense if it were something like `limitToCurrentUsers`. Maybe there's a better name, but "default query" does not give a great sense of what's actually happening under the hood.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Users.java
```diff
@@ -45,16 +45,22 @@
     private static final ObjectMapper mapper = new ObjectMapper();
     private static final Logger LOG = LoggerFactory.getLogger(Auth0Users.class);
 
+    private static URI getUrl(String searchQuery, int page, int perPage, boolean includeTotals) {
```
landonreed, line 48: Could use javadoc (can be simple though/no need to add params).

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -361,14 +408,24 @@ private static void setHeaders(Request sparkRequest, HttpRequestBase auth0Reques
...
+     * @param req The initiating request that came into datatools-server
      */
     private static String getUserIdUrl(Request req) {
+        return getUserIdUrl(req, req.params("id"));
+    }
+
+    /**
+     * Safely parse the userId and create an Auth0 url.
+     *
+     * @param req The initiating request that came into datatools-server
```
landonreed, line 420: Here (and for `getAuth0UsersExcludeDefaultQuery`) there is an extra line in the javadoc. I don't think that's the convention elsewhere in this app.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -167,24 +182,56 @@ private static String createPublicUser(Request req, Response res) {
...
-
-        JsonNode jsonNode = parseJsonFromBody(req);
         String json = String.format("{" +
                 "\"connection\": \"Username-Password-Authentication\"," +
                 "\"email\": %s," +
                 "\"password\": %s," +
                 "\"app_metadata\": {\"datatools\": [%s] } }"
                 , jsonNode.get("email"), jsonNode.get("password"), jsonNode.get("permissions"));
         setRequestEntityUsingJson(createUserRequest, json, req);
-
```
landonreed, line 184: This section of the code has always been a bit messy with all of the JSON + escaped chars. It might be worth constructing an ObjectNode to create the JSON.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -167,24 +182,56 @@ private static String createPublicUser(Request req, Response res) {
...
      */
     private static String createUser(Request req, Response res) {
+        JsonNode jsonNode = parseJsonFromBody(req);
+        String email = jsonNode.get("email").asText();
+        Auth0UserProfile user = checkForExistingAuth0User(email);
+
+        // User already exists, update permissions
+        if (user != null) {
+            String json = String.format("{ \"app_metadata\": { \"datatools\" : %s  }}", jsonNode.get("permissions"));
+            return updateUserPermissions(req, user, json);
```
landonreed, line 196: I think the datatools key should have an array value (see below at line 206).

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -395,6 +453,20 @@ private static JsonNode parseJsonFromBody(Request req) {
...
+     * Utility method to parse generic objects from JSON String and return as list.
+     * @param json The string representation of JSON to be parsed.
+     * @param clazz The class used to map each JSON object extracted.
+     */
+    private static <T> List<T> getPOJOFromJSONAsList(String json, Class<T> clazz) {
+        try {
+            JavaType type = mapper.getTypeFactory().constructCollectionType(List.class, clazz);
+            return mapper.readValue(json, type);
+        } catch (JsonProcessingException e) {
+            return null;
```
landonreed, line 466: Swallowing exceptions has caused some issues in some of the JSON parsing code in otp-middleware. It might be better just to throw the exception and let the calling method handle (and log) it.

landonreed (changes requested):
Looks good! Just a few things need some attention.

br648:
@binh-dam-ibigroup @landonreed The following two tests are failing:

canCreateUser and canReturnMeaningfulAuth0Error

Both make a call to UserController -> createUser. I have updated this to make a call to Auth0 to check if a user already exist. The response that is returned in both cases is the GET to Auth0 (check for user) and not the subsequent POST to Auth0 (create user) which is what the test is expecting. Well, this is at least my understanding!

Is this to do with creating a wiremock stub for the check user?

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Yep, had spotted that, will be in the next commit.

landonreed:
@br648, based on the Travis log output, I would say yea, you're probably right. These requests probably need corresponding wiremock stubs: https://travis-ci.org/github/ibi-group/datatools-server/builds/729898802#L10414-L10421

src/main/java/com/conveyal/datatools/manager/auth/Auth0Users.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/auth/Auth0Users.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Decided to follow the approach of the other JSON method by using the request and `logMessageAndHalt`

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Decided to follow the approach of the other JSON method by using the request and `logMessageAndHalt`

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: @landonreed there many be a reason for this. The permissions return from the UI are already raw JSON, placing this into an ObjectNode escapes the quotes and produces the following mess. I don't think there is a way around this. Where possible I have used ObjectNodes but the impact is now minimal.
```

{
  "datatools": [
    "{\"permissions\":[],\"projects\":[{\"project_id\":\"77ef6322-79b1-4c44-afc3-5d767b2b9f02\",\"permissions\":[{\"type\":\"administer-project\"}]}],\"organizations\":[],\"client_id\":\"MPXcM4ayjpQhyVKXbgQ5oXAeixpPqHd2\"}"
  ]
}
```

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Turns out that the array value [] is provided so isn't needed here. I have updated with a comment.

binh-dam-ibigroup: Updating permissions causes loss of existing permissions that were in place for that Auth0 user, so this won't work if an agency manages multiple instances (e.g. test and production) from the same Auth0 connection.
I think we want to append the permissions if it is an existing auth0 user.

binh-dam-ibigroup: Also, I pulled the latest code and the `datatools` brackets seem to be still not inserted:
image

binh-dam-ibigroup (changes requested):
I think we need to add an entry to the datatools array instead of replacing that array.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -32,7 +33,6 @@
 import org.slf4j.LoggerFactory;
 import spark.Request;
 import spark.Response;
-
```
landonreed, line 34: I think we typically include a new line between imports from dependencies and native java packages.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -42,10 +42,7 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
-import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
-
```
landonreed, line 47: New line between native package imports and static imports.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

binh-dam-ibigroup (approved):
Seems to be working well now.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -148,43 +149,158 @@ private static int getUserCount(Request req, Response res) {
      * HTTP endpoint to create a "public user" that has no permissions to access projects in the application.
      *
      * Note, this passes a "blank" app_metadata object to the newly created user, so there is no risk of someone
-     * injecting permissions somehow into the create user request.
+     * injecting permissions somehow into the create user request. If the Auth0 user already exists, update the user
+     * permissions instead.
      */
     private static String createPublicUser(Request req, Response res) {
```
landonreed, line 155: This endpoint is actually no longer used within the Data Tools application (there is no corresponding UI component), so I'm inclined to take this opportunity to remove it. Apologies for not bringing this up sooner -- after thinking about it for a bit it just doesn't belong in the application any longer.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -383,18 +495,38 @@ private static String getUserIdUrl(Request req) {
...
     }
 
+
+    /**
+     * Utility method to parse generic objects from JSON String and return as list.
+     * @param req The initiating request that came into datatools-server
+     * @param json The string representation of JSON to be parsed.
+     * @param clazz The class used to map each JSON object extracted.
+     */
+    private static <T> List<T> getPOJOFromJSONAsList(Request req, String json, Class<T> clazz) {
```
landonreed, line 520: Since this file is pretty long, I think all of these utilities might fit better in the JsonUtil file.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -154,6 +340,16 @@ public void canCreateUser() throws IOException {
      */
     @Test
     public void canReturnMeaningfulAuth0Error() throws IOException {
+
+        String url = USERS_API_PATH + "?sort=email%3A1&per_page=10&page=0&include_totals=false&search_engine=v3&q=email%3A" + URLEncoder.encode(emailForExistingAccount, "UTF-8") + "*";
```
landonreed, line 344: This common set of query params should be placed in a constant and reused.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -113,13 +116,196 @@ public void canListFirstTenUsers() throws IOException {
...
+
+        // make sure the response matches the saved snapshot
+        System.out.println(updateUserResponse.toPrettyString());
+        assertThat(updateUserResponse, matchesSnapshot());
+
+    }
+
+    /**
+     * When creating a new datatools user which matches an existing Auth0 user, make sure the permissions for the
+     * existing Auth0 user can not be updated for the same datatools instance.
```
landonreed, line 238: Should this comment (and the test name) use the word "will" instead of "can"

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -113,13 +116,196 @@ public void canListFirstTenUsers() throws IOException {
...
+        ObjectNode requestJson = mapper.createObjectNode();
+        requestJson.put("email", emailForExistingAccount);
+        ObjectNode testClientPermissions = mapper.createObjectNode();
+        testClientPermissions.put("type", "administer-application");
+        ObjectNode testClientData = mapper.createObjectNode();
+        testClientData.putArray("permissions").add(testClientPermissions);
+        testClientData.putArray("projects");
+        testClientData.putArray("organizations");
+        testClientData.put("client_id", "testing-client-id");
+        requestJson.set("permissions", testClientData);
```
landonreed, line 161: A lot of this construction of JSON looks to be repeated with other tests. Opportunity to factor this out?

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
```diff
@@ -113,13 +116,196 @@ public void canListFirstTenUsers() throws IOException {
         assertThat(userResponse, matchesSnapshot());
     }
 
+    /**
+     * When creating a new datatools user which matches an existing Auth0 user, make sure the permissions for the
+     * existing Auth0 users can be updated.
```
landonreed, line 121: When creating a new datatools user for an email address that already exists as an Auth0 user (for a different datatools client/application in Auth0), make sure..."

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -148,43 +149,158 @@ private static int getUserCount(Request req, Response res) {
...
+    }
+
+    /**
+     * Update user permissions. Only used in cases where the Auth0 user already exists and permissions for the current
+     * instance of datatools has not been set previously.
+     */
+    private static String updateUserPermissions(Request req, Auth0UserSearchResult searchResult, String newPermissions) {
+        // if there are existing permissions the assumption is that the datatools info has been set, but, only if it
+        // matches the current instance of datatools (AUTH0_CLIENT_ID).
+        if(!searchResult.existingPermissions.isEmpty() && searchResult.user.getApp_metadata().getDatatoolsInfo() != null) {
```
landonreed, line 282: I'm not entirely following the conditions where this block will evaluate to true.

landonreed:
@br648, see changes in #344 for how I might move some things around. Part of this includes removing the createPublicUser endpoint, which I should have mentioned earlier (apologies!).

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
br648: Updated.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
br648: Updated.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
br648: Updated.

src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -145,79 +141,158 @@ private static int getUserCount(Request req, Response res) {
...
+    private static Auth0UserSearchResult checkForExistingAuth0User(Request req, String email) {
+        String query = String.format("email:%s*", email);
+        String searchResult = Auth0Users.getUnrestrictedAuth0Users(query);
+        // Only ever expecting at most one user.
+        List<Auth0UserProfile> users = JsonUtil.getPOJOFromJSONAsList(req, searchResult, Auth0UserProfile.class);
+
+        if (users == null || users.isEmpty()) {
+            return null;
+        } else {
+            Auth0UserSearchResult result = new Auth0UserSearchResult();
```
landonreed, line 203: Didn't notice this before, but it seems like some of the below logic should be contained within the `Auth0UserSearchResult` constructor?

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -145,79 +141,158 @@ private static int getUserCount(Request req, Response res) {
...
+//        // preserve any previously held permissions.
+//        String permissions = (searchResult.existingPermissions.isEmpty()) ?
+//            newPermissions : String.format("%s,%s", searchResult.existingPermissions, newPermissions);
+//        String json = String.format("{ \"app_metadata\": { \"datatools\" : [%s]  }}",permissions);
+//        HttpPatch updateUserRequest = new HttpPatch(getUserIdUrl(req, searchResult.user.getUser_id()));
+//        setHeaders(req, updateUserRequest);
+//        setRequestEntityUsingJson(updateUserRequest, json, req);
+//        return executeRequestAndGetResult(updateUserRequest, req);
+//    }
+
```
landonreed, line 261: I'm not sure why the above is just commented out and not deleted.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -145,79 +141,158 @@ private static int getUserCount(Request req, Response res) {
...
+        String json = "{ \"app_metadata\": { \"datatools\" : " + data + " }}";
+        return updateUser(user, userId, json, req);
+    }
 
+    /**
+     * Perform an update user request with Auth0.
+     * @param user - the user to update
+     * @param json - the permissions (or other object) as a JSON string
+     * @param req - the original HTTP request
+     * @return the response string
```
landonreed, line 276: javadoc does not match (missing userId). Although, I'm not really sure that userId is necessary? Can't that be derived from user?

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
```diff
@@ -145,79 +141,158 @@ private static int getUserCount(Request req, Response res) {
...
-            Map.Entry<String, JsonNode> entry = fieldsIter.next();
-        }
-//        if (!data.has("client_id")) {
-//            ((ObjectNode)data).put("client_id", DataManager.config.retrieveById("auth0").retrieveById("client_id").asText());
-//        }
-        String json = "{ \"app_metadata\": { \"datatools\" : " + data + " }}";
-
+//        JsonNode jsonNode = JsonUtil.parseJsonFromBody(req);
+//        JsonNode data = jsonNode.get("data");
+//        String json = "{ \"app_metadata\": { \"datatools\" : " + data + " }}";
```
landonreed, line 294: Can this be deleted?

landonreed (approved):
Looks good. Conditional approval on removing some dead code, fixing some minor items.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: @landonreed The userId is used in the logMessageAndHalt call when the user parameter is null.

src/main/java/com/conveyal/datatools/manager/controllers/api/UserController.java
br648: Updated.

PR #359, opened 2021-01-11
Use jitpack for gtfs-lib

landonreed:
### Checklist

### Description

This PR swaps the gtfs-lib from being fetched from the maven central repo (which is an absolute pain to deploy to) to the GitHub repo using jitpack.io in anticipation of gtfs-lib ceasing publishing to maven central (primarily due to growing pains associated with switching from Travis CI to GitHub actions).


binh-dam-ibigroup (approved):

br648 (approved):
LGTM!

pom.xml
```diff
@@ -256,9 +256,9 @@
          - AWS S3 SDK - putting/getting objects into/out of S3.
          -->
         <dependency>
-            <groupId>com.conveyal</groupId>
+            <groupId>com.github.conveyal</groupId>
             <artifactId>gtfs-lib</artifactId>
-            <version>6.2.0</version>
+            <version>6.2.1</version>
```
evansiroky, line 261: Seems like this should be a commmit hash to make sure the latest in the dev branch is used.

evansiroky (changes requested):

pom.xml
landonreed: @evansiroky, I'm not sure I follow. Don't we want to reference a specific numbered version in the `pom.xml`?

evansiroky: I'm not sure what the exact behavior looks like with this. But I'd assume jitpack looks for a git tag for this version number. And 6.2.1 is only in master and the dev branch has a few commits that are ahead of this. We use the commit hash convention for the java-snapshot-matcher package already in this repo.

landonreed: Since we're intending to keep using maven-semantic-release to create tagged releases for gtfs-lib, I'm inclined to keep using those version numbers here. The numbered releases are still nice to have for release notes and for better discoverability.

PR #375, opened 2021-04-28
Consolidate AutoDeployJob initialization in ProcessSingleFeedJob

evansiroky:
This consolidates the initialization of the AutoDeployJob within the ProcessSingleFeedJob as every feed version will ultimately be processed by this job.


br648:
Looks good. Will merge.

PR #377, opened 2021-05-03
Add code to end auto-deployment in certain cases

evansiroky:
This adds some code that will prevent endless auto-deployment loops by adding 2 cases where auto-deployment will not proceed:

1. AutoDeployments won't get initiated from deploy jobs unless there is at least one feed version that completed after the last project auto-deploy time. Any feed versions yet to have their fetching/processing completed will still kick off another deploy job.
2. AutoDeployments won't start a deploy job if the updated feed version Ids for the deployment exactly match the deployment's previous feed version Ids.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
@@ -63,7 +63,9 @@ import com.conveyal.datatools.manager.models.CustomFile;
 import com.conveyal.datatools.manager.models.Deployment;
 import com.conveyal.datatools.manager.models.EC2Info;
 import com.conveyal.datatools.manager.models.EC2InstanceSummary;
+import com.conveyal.datatools.manager.models.FeedVersion;
 import com.conveyal.datatools.manager.models.OtpServer;
+import com.conveyal.datatools.manager.models.Project;
 import com.conveyal.datatools.manager.persistence.Persistence;
 import com.conveyal.datatools.manager.utils.StringUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -574,10 +576,39 @@ public void jobFinished () {
...
+
+                            // get the latest feed version (it could be null)
+                            FeedVersion latest = Persistence.feedVersions.getById(feedVersionId)
+                                    .parentFeedSource()
+                                    .retrieveLatest();
+                            // return true if the latest feed version was created after the last time an auto-deploy job
+                            // completed for the project. This means there is at least one new feed version that hasn't
+                            // yet been auto deployed that must have been created while this deploy job ran.
+                            return latest != null && latest.dateCreated.after(project.lastAutoDeploy);
+                        })
```
landonreed, line 607: This seems ripe for placing inside a method (rather than having a giant block of conditions).

landonreed (changes requested):
See comment about placing conditions into a method.

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
evansiroky: Refactored in 1c429e853fe8d0f1eae68ff5c87998a22bd75f1c

br648 (approved):
Good addition.

br648:
@evansiroky Shall I go ahead and merge?

PR #379, opened 2021-05-05
Combine TypedFeedTransformation and FeedTransformation

evansiroky:
Title mostly says it. The TypedFeedTransformation and FeedTransformation classes can be combined together.


binh-dam-ibigroup (approved):
Looks good, thanks for the suggestion!

Issue #391, opened 2021-06-15
Normalize Field substitution patterns should be validated


binh-dam-ibigroup:
## Observed behavior

Invalid regexes are accepted as search patterns for substitutions in `NormalizeFieldTransformation`
and result in failures loading GTFS feeds with a generic message "Unknown error encountered while transforming zip file", making difficult to troubleshoot.

![image](https://user-images.githubusercontent.com/56846598/122050819-dddbb100-cdb1-11eb-8ea9-b46cb5178b37.png)

## Expected behavior

A message indicating an invalid search pattern should be shown.

## Steps to reproduce the problem

1. Create a feed source
2. Add a Normalize Field Transformation (under Settings tab, Feed Transformations)
3. Under substitutions, enter an invalid substitution, e.g. `\Cir\b`.
4. Upload a new GTFS zip file and observe the error message.

## Any special notes on configuration used

MTC config

## Version of datatools-server and datatools-ui if applicable (exact commit hash or branch name)

Recent dev branch with normalize fields functionality.

PR #394, opened 2021-06-15
refactor(FeedSourceController): validate patterns on save

landonreed:
Contains some tweaks for #391 so that the patterns are validated when the feed source is saved.


binh-dam-ibigroup (changes requested):
Willing to take the approach. Please advise regarding some of the persistence errors, [see here](https://github.com/ibi-group/datatools-server/pull/394/checks?check_run_id=2832450257#step:14:12981)

binh-dam-ibigroup (approved):

PR #401, opened 2021-06-30
ci: don’t push to s3 when running on pull requests

### LINKED ISSUE #400: GH Actions: Deploy to S3 only on push (not pull_request)
## Observed behavior

When GH Actions runs on PRs, jars are uploaded to s3 on push and pull_request. However, the pull_request jar is built from a commit that does not exist in the repo (it's a temporary merge commit that is discarded). This means that tracing the commit back to the repo is met with a message on GitHub like "This commit does not exist".

From https://frontside.com/blog/2020-05-26-github-actions-pull_request/:
> This difference means that a pull_request workflow ref would look like refs/remotes/pull/##/merge whereas a push workflow would be refs/heads/branch_name. This explains why the SHA of a push workflow matches the commit that triggered the workflow, whereas the SHA of a pull_request workflow does not; instead the SHA of the pull_request is the resulting commit that was created from merging the base to the head.

## Expected behavior

We should only deploy to S3 on `push` and skip this stage for `pull_request`.

View the actions for any pull_request.

N/A

latest dev

miles-grant-ibigroup:
### Checklist

### Description

Adjusts the Github actions to not prepare for S3 deployment or or deploy to S3 when running on pull requests. Resolves #400.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -103,6 +103,7 @@ jobs:
           semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github,@conveyal/maven-semantic-release --verify-release @conveyal/maven-semantic-release --use-conveyal-workflow --dev-branch=dev --skip-maven-deploy
       # The git commands get the commit hash of the HEAD commit and the commit just before HEAD.
       - name: Prepare deploy artifacts
+        if: github.event_name != 'pull_request'
         run: |
           # get branch name of current branch for use in jar name
           export BRANCH=$GITHUB_HEAD_REF_SLUG
@@ -122,5 +123,6 @@ jobs:
           FIRST_JAR="${ALL_JARS[0]}"
           cp "$FIRST_JAR" "deploy/dt-latest-$BRANCH_CLEAN.jar"
       - name: Deploy to S3
+        if: github.event_name != 'pull_request'
         run: |
           aws s3 cp ./deploy s3://datatools-builds --recursive --acl public-read
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -107,7 +107,7 @@ jobs:
         if: github.event_name == 'push'
         run: |
           # get branch name of current branch for use in jar name
-          export BRANCH=$GITHUB_HEAD_REF_SLUG
+          export BRANCH=$GITHUB_REF_SLUG
           # Replace forward slashes with underscores in branch name.
           export BRANCH_CLEAN=${BRANCH//\//_}
           # Create directory that will contain artifacts to deploy to s3.
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -103,7 +103,8 @@ jobs:
           semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github,@conveyal/maven-semantic-release --verify-release @conveyal/maven-semantic-release --use-conveyal-workflow --dev-branch=dev --skip-maven-deploy
       # The git commands get the commit hash of the HEAD commit and the commit just before HEAD.
       - name: Prepare deploy artifacts
-        if: github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -103,6 +103,7 @@ jobs:
           semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github,@conveyal/maven-semantic-release --verify-release @conveyal/maven-semantic-release --use-conveyal-workflow --dev-branch=dev --skip-maven-deploy
       # The git commands get the commit hash of the HEAD commit and the commit just before HEAD.
       - name: Prepare deploy artifacts
+        if: github.event_name != 'pull_request'
```
landonreed, line 106: My only comment would that this can sometimes be the kind of random CI tweak that gets lost to the sands of time. So I might inject a comment here explaining why we're skipping. Also, I don't really expect us to have this workflow run on other events besides `push` and `pull_request`, but a more precise check might actually be `github.event_name == 'push'` (in case other events get added here and we don't want deploy to run).

```suggestion
        # Only deploy on push (pull_request will deploy a temp. merge commit. See #400.)
        if: github.event_name == 'push'

File: .github/workflows/maven.yml [hunk 2]
```diff
@@ -122,5 +123,6 @@ jobs:
           FIRST_JAR="${ALL_JARS[0]}"
           cp "$FIRST_JAR" "deploy/dt-latest-$BRANCH_CLEAN.jar"
       - name: Deploy to S3
+        if: github.event_name != 'pull_request'
```
landonreed, line 126:
```suggestion
        if: github.event_name == 'push'
```

landonreed (approved):
Looks good, approval is conditional on just a couple of nits.

miles-grant-ibigroup:
All makes a lot of sense thanks!

.github/workflows/maven.yml
```diff
@@ -122,5 +124,6 @@ jobs:
           FIRST_JAR="${ALL_JARS[0]}"
           cp "$FIRST_JAR" "deploy/dt-latest-$BRANCH_CLEAN.jar"
```
binh-dam-ibigroup, line 125: In the datatools-build S3 bucket, this jar appears as `dt-latest-.jar` with the same date than `dt-53e045b.jar` of this branch. `dt-latest-.jar` should probably carry the branch name (see comment above).

.github/workflows/maven.yml
```diff
@@ -103,6 +103,8 @@ jobs:
           semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github,@conveyal/maven-semantic-release --verify-release @conveyal/maven-semantic-release --use-conveyal-workflow --dev-branch=dev --skip-maven-deploy
       # The git commands get the commit hash of the HEAD commit and the commit just before HEAD.
       - name: Prepare deploy artifacts
+        # Only deploy on push (pull_request will deploy a temp. merge commit. See #400.)
+        if: github.event_name == 'push'
         run: |
           # get branch name of current branch for use in jar name
           export BRANCH=$GITHUB_HEAD_REF_SLUG
```
binh-dam-ibigroup, line 110: Interesting, I don't see a jar with the branch name in the datatools-builds S3 bucket. Are we using the wrong git variable, and maybe it should be `GITHUB_REF_SLUG`? (see https://github.com/rlespinasse/github-slug-action#slug-variables)

binh-dam-ibigroup (changes requested):
Good idea overall, see my comment about using `GITHUB_REF_SLUG` for naming one of the jars.

.github/workflows/maven.yml
miles-grant-ibigroup: Yeah seems like `GITHUB_REF_SLUG` will make more sense. I'll give it a shot and we'll see if it arrives correctly. Thanks for noticing that and letting me know!

.github/workflows/maven.yml
miles-grant-ibigroup: Screen Shot 2021-07-01 at 11 53 31 AM
This looks better right?

binh-dam-ibigroup (approved):
Thanks for the change @miles-grant-ibigroup.

Issue #123, opened 2018-12-11
Still log something when the body content is null


landonreed:
<a href="https://github.com/evansiroky"><img src="https://avatars1.githubusercontent.com/u/3112493?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [evansiroky](https://github.com/evansiroky)**
_Friday Jul 13, 2018 at 04:48 GMT_
_Originally opened as https://github.com/catalogueglobal/datatools-server/pull/104_

----

----

_**[evansiroky](https://github.com/evansiroky)** included the following code: https://github.com/catalogueglobal/datatools-server/pull/104/commits_

PR #403, opened 2021-07-15
Add Labels to Backend

miles-grant-ibigroup:
### Checklist

### Description

In preparation for adding functionality to add labels to feed sources, this PR adds labels to the server.

Labels exist and are managed like feed sources do with similar API endpoints. Feed sources can include any number of labels. Labels are attached to feed sources via update or creation (with none being the default).

There is a lot of double permission checking, as we need the request object to determine if labels can be accessed. If there is a cleaner/better way to do this please let me know!

Email notifications are disabled for label updates, checked via a convoluted equals method... I hope someone has a better idea here!

Is best used in tandem with https://github.com/ibi-group/datatools-ui/pull/693


src/main/java/com/conveyal/datatools/manager/DataManager.java
```diff
@@ -9,17 +9,7 @@
...
-import com.conveyal.datatools.manager.controllers.api.FeedSourceController;
-import com.conveyal.datatools.manager.controllers.api.FeedVersionController;
-import com.conveyal.datatools.manager.controllers.api.GtfsPlusController;
-import com.conveyal.datatools.manager.controllers.api.NoteController;
-import com.conveyal.datatools.manager.controllers.api.OrganizationController;
-import com.conveyal.datatools.manager.controllers.api.ProjectController;
-import com.conveyal.datatools.manager.controllers.api.ServerController;
-import com.conveyal.datatools.manager.controllers.api.StatusController;
-import com.conveyal.datatools.manager.controllers.api.UserController;
+import com.conveyal.datatools.manager.controllers.api.*;
```
evansiroky, line 12: I think we have a preference for listing out all these dependencies one-by-one, but I don't think we have a standardized IDE formatting file. Maybe we should though.

src/main/java/com/conveyal/datatools/manager/models/Project.java
```diff
@@ -105,6 +105,19 @@ public Project() {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Get all the labels for this project.
+     */
+    public Collection<Label> retrieveProjectLabels() {
+        return Persistence.labels.getAll().stream()
+                .filter(label -> this.id.equals(label.projectId))
+                .collect(Collectors.toList());
```
evansiroky, line 114: Please make a query to the database that does the filtering within the database so that not all labels are fetched from the database.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        if (!authorized) {
+            // Special message if viewing an adminOnly label
+            if (label.adminOnly) {
+                logMessageAndHalt(req, 403, "User is not admin so cannot view admin-only label");
+            }
+            // Throw halt if user not authorized.
+            logMessageAndHalt(req, 403, "User is not admin so cannot update or create labels");
+        }
```
evansiroky, line 211: These `logMessageAndHalt`s could be handled within the above switch statement. Also, the `if(label.adminOnly)` could also apply to the `CREATE`, `MANAGE` or `DELETE` cases, so the message may not be correct anyways.

```suggestion
        boolean isProjectAdmin = userProfile.canAdministerProject(label.projectId, label.organizationId());
        switch (action) {
            case CREATE:
            case MANAGE:
            case EDIT:
                // Only project admins can edit labels
                logMessageAndHalt(req, 403, "User is not admin so cannot update or create labels");
                break;
            case VIEW:
                if (label.adminOnly && !isProjectAdmin) {
                    logMessageAndHalt(req, 403, "User is not admin so cannot view admin-only label");
                }
                break;
        }
```

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,224 @@
...
+        String projectId = req.queryParams("projectId");
+        Project project = Persistence.projects.getById(projectId);
+        if (project == null) {
+            logMessageAndHalt(req, 400, "Must provide valid projectId query param to retrieve labels.");
+        }
+
+        Collection<Label> projectLabels = project.retrieveProjectLabels();
+        for (Label label: projectLabels) {
+            // Only return labels user has access to.
+            labelsToReturn.add(checkLabelPermissions(req, label, Actions.VIEW));
```
evansiroky, line 57: This could end up inserting null labels. Please refactor so that only non-null values are returned.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /** Helper method returns label only if user is authorized to do given action with it
+     *
+     * @param req       Spark request used for determining user permissions
+     * @param label     Label to check
+     * @param action    Action to be taken on label, changes who can do what
+     * @return          Label which is ok to return to user
+     */
+    public static Label checkLabelPermissions(Request req, Label label, Actions action) {
```
evansiroky, line 176: It seems the primary goal of this method is to call `logMessageAndHalt` and not do anything with the label. Therefore, this method would probably be better if it had a void return type.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,224 @@
...
+
+        try {
+            Persistence.labels.create(newLabel);
+            return newLabel;
+        } catch (Exception e) {
+            logMessageAndHalt(req, 500, "Unknown error encountered creating label", e);
+            return null;
+        }
+    }
+    /**
```
evansiroky, line 80: Please add empty line in-between.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
evansiroky: Actually, this current design would end up halting the request instead of returning non-admin-viewable labels.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -141,6 +139,11 @@ private static void validate(Request req, FeedSource feedSource) {
         if (feedSource.retrieveProject() == null) {
             validationIssues.add("Valid project ID must be provided.");
         }
+        for (String labelId: feedSource.labels) {
+            if (Persistence.labels.getById(labelId) == null) {
```
evansiroky, line 143: It's a little inefficient to be making multiple individual queries for a single label like this.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -332,10 +343,22 @@ public static FeedSource checkFeedSourcePermissions(Request req, FeedSource feed
             // Throw halt if user not authorized.
             logMessageAndHalt(req, 403, "User not authorized to perform action on feed source");
         }
+
+        // Remove labels user is not allowed to see if user is not admin
+        if (!isAdmin) {
+            feedSource.labels = feedSource.labels.stream()
+                    // Need to resolve label IDs to labels, then back
+                    .map(labelId -> Persistence.labels.getById(labelId))
```
evansiroky, line 351: It's a little inefficient to be making multiple individual queries for a single label like this.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -17,16 +17,15 @@
...
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonView;
+import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.model.Sorts;
 import org.bson.codecs.pojo.annotations.BsonIgnore;
+import org.bson.codecs.pojo.annotations.BsonProperty;
```
evansiroky, line 28: This is not used in this file.

src/main/java/com/conveyal/datatools/manager/serializers/LabelSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    protected LabelSerializer(JavaType type) {
+        super(type);
+    }
+
+    protected LabelSerializer(Class<?> t, boolean dummy) {
+        super(t, dummy);
+    }
+
+    protected LabelSerializer(StdSerializer<?> src) {
```
evansiroky, line 39: Please add JSDoc for what each of these do.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -78,6 +74,8 @@ private static FeedSource getFeedSource(Request req, Response res) {
         }
         Collection<FeedSource> projectFeedSources = project.retrieveProjectFeedSources();
         for (FeedSource source: projectFeedSources) {
+            // Verify user permissions before returning source
+            source = checkFeedSourcePermissions(req, source, Actions.VIEW);
```
evansiroky, line 78: Isn't this already handled with some code below?

evansiroky (changes requested):
See various comments.

Also, one big change I'd like to request is to change the database/model schema a bit here as it pertains to the relationship between the labels and feed sources. Currently, there is a number of code paths that make multiple "requestById" for individual labels. Also, there is some nice magic going on that parses/outputs the labels of a feed source. In order to reduce this action, I propose to remove the `labels` field from the FeedSource model and instead create a `List<String> feedSourceIds` field within the Label model. I believe doing this would simplify the backend code a fair amount: there would no longer be a need for the `LabelSerializer` class and there shouldn't be any need to change the `FeedSource` model or the `FeedSourceController`.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Disregard if refactoring as suggested in https://github.com/ibi-group/datatools-server/pull/403#pullrequestreview-714190079.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Disregard if refactoring as suggested in https://github.com/ibi-group/datatools-server/pull/403#pullrequestreview-714190079.

src/main/java/com/conveyal/datatools/manager/serializers/LabelSerializer.java
evansiroky: Disregard if refactoring as suggested in https://github.com/ibi-group/datatools-server/pull/403#pullrequestreview-714190079.

src/main/java/com/conveyal/datatools/manager/DataManager.java
miles-grant-ibigroup: Yes IntelliJ did this for me on save, but I've disabled this now. Reverted in 
85261da

src/main/java/com/conveyal/datatools/manager/models/Project.java
miles-grant-ibigroup: Good fix and is fewer lines as well! Have also made this change to filtering feed sources in 
85261da

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: This looks much cleaner thank you! Committed in 
85261da

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: This is a good catch thank you! Resolved in 
85261da

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Was architecting this based on FeedSource, but maybe a little too closely! Have fixed this in 
85261da

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Added in 
85261da

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: Removed in 85261da

miles-grant-ibigroup:
Thank you for all your comments! Have touched everything except the FeedSourceController until we figure out the best course of action there.

I agree the serializer isn’t ideal. The idea behind adding labels to feed sources was to both reduce the complexity of the changes needed in datatools-ui, as well as reducing the number of network requests when rendering the dashboard. However, I agree that things could be a bit cleaner. 

One option that might be the best of both worlds could be to change the back-end to assign feed sources to labels, but include a custom method in feed sources that goes through all labels and returns those in the feed source? For updating, the front-end could either be re-written to make a separate call to update each label, or a custom serializer could handle this? Curious about your thoughts. To me, the most important aspect is ensuring the number of network requests doesn't increase, but that may be misguided!

landonreed (changes requested):
I really like the approach here. Just a couple of changes/suggestions from me.

I'm kind of torn on whether to go with `Label#feedSourceIds` or `FeedSource#labels`. I might be starting to lean to the first option, but my decision somewhat depends on a question for @miles-grant-ibigroup: Does the serialize `List<Label> labels` attached to FeedSource include both admin and non-admin labels? If so, a sneaky non-admin user could inspect the network response and see what admin labels exist. This is a pretty minor concern overall, but it does effectively defeat some of the guards we're putting in place.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -190,6 +197,11 @@ private static FeedSource updateFeedSource(Request req, Response res) throws IOE
...
         }
         Persistence.feedSources.replace(feedSourceId, updatedFeedSource);
+
+        if (formerFeedSource.equalsExceptLabels(updatedFeedSource)) {
+            // If only labels have changed, don't send out an email
+            return updatedFeedSource;
+        }
         // After successful save, handle auto fetch job setup.
         Scheduler.handleAutoFeedFetch(updatedFeedSource);
         // Notify feed- and project-subscribed users after successful save
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good idea! This is much more expandable. I've made the syntax inviting further filters in cf247b0

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,220 @@
...
+     * Spark HTTP endpoint to get a single label by ID
+     */
+    private static Label getLabel(Request req, Response res) {
+        return requestLabelById(req, Actions.VIEW);
+    }
+
+    /**
+     * Spark HTTP endpoint that handles getting all labels for a handful of use cases:
+     * - for a single project (if projectId query param provided)
+     * - for the entire application
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch thank you! Removed in cf247b0

src/main/java/com/conveyal/datatools/manager/serializers/LabelSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+import java.sql.Array;
+import java.util.ArrayList;
+
+public class LabelSerializer extends StdSerializer<ArrayList<String>> {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Have added clarifications in 
cf247b0

src/main/java/com/conveyal/datatools/manager/serializers/LabelSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+        for (String labelId: strings) {
+            jsonGenerator.writeStartObject();
+
+            Label label = Persistence.labels.getById(labelId);
+            jsonGenerator.writeStringField("id", label.id);
+            jsonGenerator.writeStringField("projectId", label.projectId);
+            jsonGenerator.writeStringField("name", label.name);
+            jsonGenerator.writeStringField("description", label.description);
+            jsonGenerator.writeStringField("color", label.color);
+            jsonGenerator.writeBooleanField("adminOnly", label.adminOnly);
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Absolutely not! Thank you for making me take another look at it. Much cleaner now cf247b0

src/main/java/com/conveyal/datatools/manager/models/Project.java
```diff
@@ -100,11 +100,20 @@ public Project() {
...
 
+    /**
+     * Get all the labels for this project.
+     */
+    public Collection<Label> retrieveProjectLabels() {
+        return Persistence.labels.getFiltered(eq("projectId", this.id));
+    }
+
+    // Keep an empty collection here which is filled dynamically later
+    public Collection<Label> labels;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Makes sense! Added in cf247b0

src/main/java/com/conveyal/datatools/manager/serializers/LabelSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Thank you! Leftover from when I was figuring out the best way to architect things. Removed in cf247b0

miles-grant-ibigroup:
Thanks Landon have made these changes, aside from the `FeedSourceController` stuff.

The serializer does return both admin and non-admin labels, but `FeedSourceController` filters the label string list before creating any response so only "correct" labels are serialized! Everything going across the network is correct based on permissions. If we went for Label#feedSourceIds, we’d have to do the same on-server filtering, just in the new endpoint/method we use to return labels.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,219 @@
...
+                }
+                break;
+            default:
+                // Incorrect query, so fail
+                logMessageAndHalt(req, 400, "Not enough information supplied to determine label access");
+                break;
+        }
+
+        // If we make it here, user has permission and the requested label is valid
+        return;
```
evansiroky, line 209: This doesn't seem needed anymore

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,219 @@
...
+        return label;
+    }
+
+    /** Helper method returns label only if user is authorized to do given action with it
+     *
+     * @param req       Spark request used for determining user permissions
+     * @param label     Label to check
+     * @param action    Action to be taken on label, changes who can do what
+     * @return          Label which is ok to return to user
+     */
```
evansiroky, line 179: Please change JavaDoc to remove "returns" wording.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,219 @@
...
+                List<String> newLabels = feedSource.labels.stream()
+                        .filter(l -> !l.equals(label.id))
+                        .collect(Collectors.toList());
+                // If there are changes, update the feed source
+                if (!newLabels.equals(feedSource.labels)) {
+                    FeedSource fs = Persistence.feedSources.getById(feedSource.id);
+                    fs.labels = newLabels;
+                    Persistence.feedSources.replace(fs.id, fs);
+                }
+            });
```
evansiroky, line 146: This is one of the big reasons I'm in the camp of `Label#feedSourceIds`. This involves potentially dozens of individual database updates to remove the label from a feed source. That is risky and inefficient to be making so many requests because if one of them fails, then there could be a situation where a bunch of the records still have the label entry in them. As an aside, the first statement that fetches all feed sources does so regardless of the project associated with the label.

src/main/java/com/conveyal/datatools/manager/persistence/TypedPersistence.java
```diff
@@ -142,6 +143,15 @@ public T getById (String id) {
         return mongoCollection.find().into(new ArrayList<>());
     }
 
+    /**
+     * Gets all objects with one of the IDs passed
+     * @param ids   List of IDs to return objects with
+     * @return      Objects in the collection with given IDs
+     */
+    public List<T> getByIds (List<String> ids) {
```
evansiroky, line 151: Is this used anywhere?

evansiroky:
> Thanks Landon have made these changes, aside from the `FeedSourceController` stuff.
> 
> The serializer does return both admin and non-admin labels, but `FeedSourceController` filters the label string list before creating any response so only "correct" labels are serialized! Everything going across the network is correct based on permissions. If we went for `Label#feedSourceIds`, we’d have to do the same on-server filtering, just in the new endpoint/method we use to return labels.

This PR already adds all of a project's labels to the Project model in the `ProjectController#checkProjectPermissions` method, so a new endpoint is not needed and all of the project's label data viewable according to admin permissions is already outputted in the ProjectController responses. See example:

Screen Shot 2021-07-27 at 3 37 32 PM

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -91,6 +105,23 @@ public void createFeedSourceWithUrlTest() {
         assertEquals(1, jobCountForFeed(feedSourceWithUrl.id));
     }
 
+    /**
+     * Create some labels, add them to the feed source make them admin only, and check that they don't appear if not an admin
+     */
+    @Test
+    public void createFeedSourceWithLabels() {
+        // Testing the following (important) components of label is as of yet not supported.
```
evansiroky, line 113: If this method ends up never having any content, then the whole method should be commented out.

evansiroky (changes requested):
I still recommend the `Label#feedSourceIds` approach for various reasons mentioned.

Also, the `Project#delete` method needs to delete all labels associated with the project being deleted.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: You're right. Removed!

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Good catch thank you. Removed!

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: I spent some time thinking about how to refactor this but am starting to come around to your side. The only other option would be to ignore non-existent labels, but that doesn't seem very smart.

src/main/java/com/conveyal/datatools/manager/persistence/TypedPersistence.java
miles-grant-ibigroup: I think it might be once we switch to `Labels#feedSourceId`, but will leave this open as a reminder in case we don't use it

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
miles-grant-ibigroup: If we switch to `Labels#feedSourceId` then I will add tests

miles-grant-ibigroup:
Have amended `Project#delete`

@evansiroky I think I've come around to doing `Label#feedSourceId`, since we're already doing a second request to get all project info on the dashboard. With a few helper methods the changes to the UI should be minimal as well.

The only issue I foresee is the assigner making a lot of requests as it has to update each label individually (which might overwhelm the server?). Do you have any thoughts on how to do this better? 

Also, wouldn't deleting a feed source still require going through all labels? Since there are generally more labels than feed sources, wouldn't this result in more database write operations?

src/main/java/com/conveyal/datatools/manager/models/Project.java
```diff
@@ -100,11 +101,21 @@ public Project() {
      */
     public Collection<FeedSource> retrieveProjectFeedSources() {
         // TODO: use index, but not important for now because we generally only have one FeedCollection
-        return Persistence.feedSources.getAll().stream()
-                .filter(fs -> this.id.equals(fs.projectId))
-                .collect(Collectors.toList());
+        return Persistence.feedSources.getFiltered(eq("projectId", this.id));
```
evansiroky, line 104: 👍

evansiroky:
> Have amended `Project#delete`
> 
> @evansiroky I think I've come around to doing `Label#feedSourceId`, since we're already doing a second request to get all project info on the dashboard. With a few helper methods the changes to the UI should be minimal as well.
> 
> The only issue I foresee is the assigner making a lot of requests as it has to update each label individually (which might overwhelm the server?). Do you have any thoughts on how to do this better?
> 
> Also, wouldn't deleting a feed source still require going through all labels? Since there are generally more labels than feed sources, wouldn't this result in more database write operations?

I chatted with Miles about these items. The potential for making multiple requests when adding multiple labels to a feed source at once is going to be something that would involve making more http requests. Also, deleting a feed source would involve doing multiple updates in multiple label records. One other option we haven't considered yet is `FeedSource#labelIds` without `LabelSerializer` as opposed to `FeedSource#labels` with `LabelSerializer`. That route would have the advantage of not having to make any database requests to return Label objects within each feed source. I have gone ahead and made this comparison table with each of these options and tried to quantify how many HTTP vs database requests would happen in each:

<table>
  <tr>
    <td rowspan=2>Type of request</td>
    <td colspan=2>FeedSource#labels (with LabelSerializer)</td>
    <td colspan=2>FeedSource#labelIds (no LabelSerializer)</td>
    <td colspan=2>Label#feedSourceIds</td>
  </tr>
  <tr>
    <td>Http Requests</td>
    <td>Database Requests</td>
    <td>Http Requests</td>
    <td>Database Requests</td>
    <td>Http Requests</td>
    <td>Database Requests</td>
  </tr>
  <tr>
    <td>Get project</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
  </tr>
  <tr>
    <td>Delete project</td>
    <td>1</td>
    <td># of labels in project</td>
    <td>1</td>
    <td># of labels in project</td>
    <td>1</td>
    <td># of labels in project</td>
  </tr>
  <tr>
    <td>Get feed source</td>
    <td>1</td>
    <td># of feed sources X # of labels on each feed source</td>
    <td>1</td>
    <td>0 (labels received from Project)</td>
    <td>1</td>
    <td>0 (labels received from Project)</td>
  </tr>
  <tr>
    <td>Update labels for feed source</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
    <td>1</td>
    <td># of labels to be updated</td>
    <td># of labels to be updated</td>
  </tr>
  <tr>
    <td>Delete feed source</td>
    <td>1</td>
    <td>0</td>
    <td>1</td>
    <td>0</td>
    <td>1</td>
    <td>fetch all labels + # of labels associated with feed source</td>
  </tr>
  <tr>
    <td>Delete label</td>
    <td>1</td>
    <td>fetch all project feed sources + # of feed sources with label</td>
    <td>1</td>
    <td>fetch all project feed sources + # of feed sources with label</td>
    <td>1</td>
    <td>1</td>
  </tr>
</table>

As can be seen, the `Label#feedSourceIds` wins if we expect both 1) that users will typically only add one label at a time to a feed source and 2) that deleting a feed source is rarer than deleting a label. Otherwise, `FeedSource#labelIds` without `LabelSerializer` wins.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: The issue was removing admin-only labels, but I've extracted this code so these lines are now removed.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: I take back my coming around to your side -- if we delete a feed source when doing `Label#feedSourceIds` we have the same problem and the same number of db calls

src/main/java/com/conveyal/datatools/manager/persistence/TypedPersistence.java
miles-grant-ibigroup: This is now used to reduce the number of stream operations

miles-grant-ibigroup:
Have now implemented the FeedSource#labelIds (no LabelSerializer) approach. I think it is the best of both worlds... If this gets the all clear I'll add some tests for all this, although I've been struggling to see how to examine the contents of an HTTP response made with `TestUtils.makeRequest()`?

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,218 @@
...
+        if (StringUtils.isEmpty(label.name)) {
+            validationIssues.add("Name field must not be empty.");
+        }
+        if (!validationIssues.isEmpty()) {
+            logMessageAndHalt(
+                    req,
+                    HttpStatus.BAD_REQUEST_400,
+                    "Request was invalid for the following reasons: " + String.join(", ", validationIssues)
+            );
+        }
```
br648, line 98: Because only a single label is being checked, I think this method can be simplified to something like:     

```
        // Label is quite forgiving (sets defaults if null) and the boolean value is type checked,
        // so there is little to validate.
        if (StringUtils.isEmpty(label.name)) {
            logMessageAndHalt(
                req,
                HttpStatus.BAD_REQUEST_400,
                "Request was invalid, the name field must not be empty."
            );
        }

```

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,218 @@
...
+        Label newLabel = getPOJOFromRequestBody(req, Label.class);
+        validate(req, newLabel);
+        // User may not be allowed to create a new label
+        checkLabelPermissions(req, newLabel, Actions.CREATE);
+
+        try {
+            Persistence.labels.create(newLabel);
+            return newLabel;
+        } catch (Exception e) {
+            logMessageAndHalt(req, 500, "Unknown error encountered creating label", e);
```
br648, line 76: Be consistant with your full stops at the end of these messages. Some have it, some don't. :-)

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,218 @@
...
+            logMessageAndHalt(req, 400, "Label ID does not exist");
+            return;
+        }
+
+        boolean isProjectAdmin = userProfile.canAdministerProject(label.projectId, label.organizationId());
+        switch (action) {
+            case CREATE:
+            case MANAGE:
+            case EDIT:
+                // Only project admins can edit labels
```
br648, line 193: Should this say:

// Only project admins can create, manage and edit labels

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,218 @@
...
+                if (!isProjectAdmin) {
+                    logMessageAndHalt(req, 403, "User is not admin so cannot update or create labels");
+                }
+            case VIEW:
```
br648, line 197: Should there be a break here? Else VIEW will always get evaluated if action is CREATE, MANAGE or EDIT.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -322,6 +328,24 @@ public int compareTo(FeedSource o) {
         return this.name.compareTo(o.name);
     }
 
+    public boolean equalsExceptLabels(FeedSource o) {
```
br648, line 331: Method comment.

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -119,6 +150,14 @@ private static FeedSource createFeedSource(String name, URL url) {
         return feedSource;
     }
 
+    /**
+     * Helper method to create label
+     */
+    private static Label createLabel(String name) {
+        Label label = new Label(name, "A label used during testing", "#123", false, project.id);
```
br648, line 157: Can be inlined. e.g.:

`        return new Label(name, "A label used during testing", "#123", false, project.id);
`

br648 (changes requested):
LGTM so far, just some minor comments from me to address.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: This is much cleaner thank you!

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Good catch thank you!

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Very good catch thank you!

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: Added

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
miles-grant-ibigroup: Who knows how this happened thanks!

miles-grant-ibigroup:
Thanks Rob for the feedback, it's addressed.

I've also added some basic tests. I'd like to expand them (to check easily-breakable admin only stuff), but can't figure `TestUtils` or `HTTPUtils` out enough to check HTTP response bodies.

miles-grant-ibigroup:
Following the new PR policy, have unassigned @landonreed and @evansiroky until @br648's review is in

br648 (changes requested):
> I've also added some basic tests. I'd like to expand them (to check easily-breakable admin only stuff), but can't figure `TestUtils` or `HTTPUtils` out enough to check HTTP response bodies.

@miles-grant-ibigroup Something like `String body = EntityUtils.toString(response.getEntity());` should get you want you need. You can search for `EntityUtils` for examples.

Looking good so far! Assign back to me once your addtional response body tests are in place.

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -7,14 +7,18 @@
 import com.conveyal.datatools.manager.models.FeedRetrievalMethod;
 import com.conveyal.datatools.manager.models.FeedSource;
 import com.conveyal.datatools.manager.models.FetchFrequency;
+import com.conveyal.datatools.manager.models.Label;
 import com.conveyal.datatools.manager.models.Project;
 import com.conveyal.datatools.manager.persistence.Persistence;
 import com.conveyal.datatools.manager.utils.HttpUtils;
 import com.conveyal.datatools.manager.utils.json.JsonUtil;
+import org.apache.http.HttpEntity;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This is cool thanks!

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -104,6 +123,67 @@ public void createFeedSourceWithNoUrlTest() {
...
+        assertEquals(OK_200, createFeedSourceResponse.getStatusLine().getStatusCode());
+        assertEquals(2, labelCountForFeed(feedSourceWithLabels.id));
+        assertEquals(2, Persistence.projects.getById(feedSourceWithLabels.retrieveProject().id).retrieveProjectLabels().size());
+
+        // Test that an admin can see 2 of them
+        // Test that a non-admin can only one of them
+        // Test that feed source output is correct
+        // Test that the feed source output only shows 1 label for a non-admin
+        // Test that a non-admin sees only 1 label in the Project output
+        // Test that the Project output contains the full Label objects
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Have added a comment clarifying what those tests are!

miles-grant-ibigroup:
I don't want upgrading HttpUtils to match otp-middleware to be part of this PR, so have removed those tests from here. They'll be moved to datatools-ui if they're implemented. Have cleaned up here for now.

br648 (approved):
Just three minor items, so approving on condition these are addressed. Good work btw!

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,214 @@
...
+import spark.Response;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.conveyal.datatools.common.utils.SparkUtils.getPOJOFromRequestBody;
+import static com.conveyal.datatools.common.utils.SparkUtils.logMessageAndHalt;
+import static spark.Spark.*;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: One day I'm going to have 1000 imports and be so confused as to why it keeps auto formatting

miles-grant-ibigroup:
Thanks @br648! Have got that all sorted. Up next is @evansiroky

src/main/java/com/conveyal/datatools/manager/models/Label.java
```diff
@@ -0,0 +1,89 @@
...
+    public Project retrieveProject() {
+        return projectId != null ? Persistence.projects.getById(projectId) : null;
+    }
+
+    @JsonProperty("organizationId")
+    @JsonIgnore
+    public String organizationId () {
+        Project project = retrieveProject();
+        return project == null ? null : project.organizationId;
+    }
```
evansiroky, line 32: Seems like the `retrieveProject` method contents could be inlined in here since that appears to be the only place it is used.

src/main/java/com/conveyal/datatools/manager/models/Label.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            Persistence.labels.removeById(this.id);
+        } catch (Exception e) {
+            LOG.error("Could not delete label", e);
+        }
+    }
+
+    public int compareTo(Label o) {
+        return this.name.compareTo(o.name);
+    }
```
evansiroky, line 84: I don't think this is used anywhere, so if it's not needed maybe it can be removed.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -144,6 +145,11 @@ public String organizationId () {
      */
     public String snapshotVersion;
 
+    /**
+     * IDs of Labels assigned to this Feed
+     */
+    public List<String> labels = new ArrayList<>();
```
evansiroky, line 151: Please rename:

```suggestion
    public List<String> labelIds = new ArrayList<>();
```

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -332,7 +344,40 @@ public static FeedSource checkFeedSourcePermissions(Request req, FeedSource feed
...
+     *
+     * @param formerFeedSource  A feed source object, without new changes
+     * @param updatedFeedSource A feed source object, with new changes
+     * @return                  A boolean value indicating if the updated feed source is changed enough to warrant sending a notification.
+     */
+    private static boolean shouldNotifyUsersOnFeedUpdated(FeedSource formerFeedSource, FeedSource updatedFeedSource) {
+        return
+                // If only labels have changed, don't send out an email
+                formerFeedSource.equalsExceptLabels(updatedFeedSource);
+    }
```
evansiroky, line 364: Just double-checking here, is this "do not notify of feed source changes when only labels change" a kind of functionality we want to enable? Perhaps some users might want to know if a label is added to a feed source.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -311,9 +322,10 @@ public static FeedSource checkFeedSourcePermissions(Request req, FeedSource feed
         }
         String orgId = feedSource.organizationId();
         boolean authorized;
+
         switch (action) {
             case CREATE:
-                authorized = userProfile.canAdministerProject(feedSource.projectId, orgId);
+                authorized = userProfile.canAdministerProject(feedSource.id, orgId);
```
evansiroky, line 328: This seems like it should not have been changed.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -332,7 +344,40 @@ public static FeedSource checkFeedSourcePermissions(Request req, FeedSource feed
...
+     * @param feedSource    The feed source to clean
+     * @param userProfile   The user object to check permissions with
+     * @return              A feed source containing only labels the user is allowed to see
+     */
+    private static FeedSource cleanFeedSourceLabels(FeedSource feedSource, Auth0UserProfile userProfile) {
+        // Remove labels user is not allowed to see if user is not admin
+        boolean isAdmin = userProfile.canAdministerProject(feedSource.id, feedSource.organizationId());
+        if (!isAdmin) {
+            feedSource.labels = Persistence.labels.getByIds(feedSource.labels).stream()
+                    .filter(label -> !label.adminOnly)
```
evansiroky, line 377: I don't have very strong feelings about this since it won't be very many records, but it seems like filtering out the `adminOnly` labels could be something done as part of the database query to ensure that the minimum required data is transferred over the network from the db to this application.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,217 @@
...
+                logMessageAndHalt(req, 400, "Not enough information supplied to determine label access.");
+                break;
+        }
+
+        // If we make it here, user has permission and the requested label is valid
+    }
+
+    public static void register (String apiPrefix) {
+        get(apiPrefix + "secure/label/:id", LabelController::getLabel, json::write);
+        get(apiPrefix + "secure/label", LabelController::getAllLabels, json::write);
```
evansiroky, line 212: Is this endpoint still needed?

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,217 @@
...
+
+        // More specific error message if label doesn't exist
+        if (label == null) {
+            logMessageAndHalt(req, 403, "Label does not exist.");
+            return null;
+        }
+
+        try {
+            // Iterate over feed sources and remove any references to the label
+            Persistence.feedSources.getAll().forEach(feedSource -> {
```
evansiroky, line 134: Please fetch just the feed sources according to the label's project ID from the database.

src/main/java/com/conveyal/datatools/manager/models/Label.java
```diff
@@ -0,0 +1,89 @@
...
+        this.name = name;
+        this.description = description != null ? description : "";
+        this.color = color != null ? color : "#000000";
+        this.adminOnly = adminOnly;
+
+        this.projectId = projectId;
+    }
+
+    /**
+     * No-arg constructor to yield an uninitialized feed source, for dump/restore.
```
evansiroky, line 64:
```suggestion
     * No-arg constructor to yield an uninitialized label, for dump/restore.
```

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -104,10 +130,62 @@ public void createFeedSourceWithNoUrlTest() {
...
+
+        // Create feed source with labels
+        HttpResponse createFeedSourceResponse = TestUtils.makeRequest("/api/manager/secure/feedsource",
+                JsonUtil.toJson(feedSourceWithLabels),
+                HttpUtils.REQUEST_METHOD.POST
+        );
+        assertEquals(OK_200, createFeedSourceResponse.getStatusLine().getStatusCode());
+        // Test that they are assigned properly
+        assertEquals(2, labelCountForFeed(feedSourceWithLabels.id));
+        assertEquals(2, Persistence.projects.getById(feedSourceWithLabels.retrieveProject().id).retrieveProjectLabels().size());
```
evansiroky, line 165: Can you also make this into a helper method:

```suggestion
        assertEquals(2, labelCountForFeedSourceProject(feedSourceWithLabels));
```

evansiroky (changes requested):
See comments

src/main/java/com/conveyal/datatools/manager/models/Label.java
miles-grant-ibigroup: Inlined in 9ac824a

src/main/java/com/conveyal/datatools/manager/models/Label.java
miles-grant-ibigroup: Yep this is no longer used. Removed in 9ac824a

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: Renamed in 9ac824a

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: I think the number of emails this would generate would be really noisy. There are some other examples throughout the code of this feature being disabled, especially since the email doesn't say which part of a feed source was updated, it just says "feed source property changed"
Screen Shot 2021-08-03 at 8 06 34 AM

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: Do not know how that happened! Fixed in 9ac824a

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: Have refactored this in both fa4ba57 and then further in 9ac824a. I think it's more efficient, but would appreciate your double-check

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: It isn't presently used by the UI, but I thought it might be worth keeping around to make future work easier? But do not have strong opinion on that so happy to remove if you think it's a better idea to remove it

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Have refactored this in both fa4ba57 and then further in 9ac824a. I think it's more efficient, but would appreciate your double-check

src/main/java/com/conveyal/datatools/manager/models/Label.java
miles-grant-ibigroup: Good catch! Thought I renamed all of those... Fixed in 9ac824a

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
miles-grant-ibigroup: Refactored in 9ac824a

miles-grant-ibigroup:
Thanks for these, have addressed all or left comments.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,234 @@
...
+     * persisted to the database.
+     */
+    private static void validate(Request req, Label label) {
+        // Label is quite forgiving (sets defaults if null) and the boolean value is type checked,
+        // so there is little to validate.
+        if (label.name.length() > 25) {
+            logMessageAndHalt(
+                    req,
+                    HttpStatus.BAD_REQUEST_400,
+                    "Request was invalid, the name was too long."
```
evansiroky, line 94: Please improve this error message to say what the maximum length is.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
```diff
@@ -0,0 +1,234 @@
...
+            );
+        }
+        if (label.description.length() > 50) {
+            logMessageAndHalt(
+                    req,
+                    HttpStatus.BAD_REQUEST_400,
+                    "Request was invalid, the description was too long."
```
evansiroky, line 101: Please improve this error message to say what the maximum length is.

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
evansiroky: In general I think it's good to remove unused code, but I'll let the final reviewer have the final say.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Yes, I believe this is noisy and I would do this, but I'm more just wanting to double check if a client is asking for this functionality to be enabled/disabled.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: I think it'd be a good idea to make one trip to the database here still by fetching using the `in` clause.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -24,6 +24,7 @@
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonView;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
```
evansiroky, line 27: This import is not used in this file.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: I checked internally with the team and am ok marking this resolved.

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -104,10 +133,71 @@ public void createFeedSourceWithNoUrlTest() {
...
+
+        // Test that they are assigned properly
+        assertEquals(2, labelCountForFeed(feedSourceWithLabels.id));
+        // Test that project shows only correct labels based on user auth
+        assertEquals(2, labelCountforProject(feedSourceWithLabels.projectId, true));
+        assertEquals(1, labelCountforProject(feedSourceWithLabels.projectId, false));
+
+        // Test that feed source shows only correct labels based on user auth
+        List<String> labelsSeenByAdmin = FeedSourceController.cleanFeedSourceLabels(feedSourceWithLabels, true).labelIds;
+        List<String> labelsSeenByViewOnlyUser = FeedSourceController.cleanFeedSourceLabels(feedSourceWithLabels, false).labelIds;
```
evansiroky, line 175: It'd be nice to send off an actual HTTP request and then parse the response here. In doing things that way, we would be testing the overall API functionality and not tying things to the internal workings of the FeedSourceController.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
evansiroky: Looks much better. Thanks!

evansiroky (changes requested):
Just a few more little things as noted since my last review.

Also, please work on one more opportunity to reduce the number of trips to the database: see https://github.com/ibi-group/datatools-server/pull/403#discussion_r682766119.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: Refactored in 4ac2e4cab7a40d0c40960ba15503c2781adef26d

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Added in 4ac2e4cab7a40d0c40960ba15503c2781adef26d

src/main/java/com/conveyal/datatools/manager/controllers/api/LabelController.java
miles-grant-ibigroup: Added in 4ac2e4cab7a40d0c40960ba15503c2781adef26d

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: Removed in 4ac2e4cab7a40d0c40960ba15503c2781adef26d

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
miles-grant-ibigroup: I agree. However, I think Landon and I agreed that the work required to mock the authorization bearers (from either the client side or the server side tests) would not be worth it at this time

miles-grant-ibigroup:
Thanks have made adjustments. For the `feedSourceController` validation, I think I managed to reduce database queries quite a lot, is that what you had in mind?

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -141,6 +149,11 @@ private static void validate(Request req, FeedSource feedSource) {
         if (feedSource.retrieveProject() == null) {
             validationIssues.add("Valid project ID must be provided.");
         }
+        for (String labelId: feedSource.labelIds) {
+            if (Persistence.labels.getByIds(feedSource.labelIds).size() != feedSource.labelIds.size()) {
+                validationIssues.add("All labels assigned to feed must exist.");
+            }
+        }
```
evansiroky, line 156: This for loop is now unneccessary. It can be rewritten as follows:

```suggestion
        if (Persistence.labels.getByIds(feedSource.labelIds).size() != feedSource.labelIds.size()) {
            validationIssues.add("All labels assigned to feed must exist.");
        }
```

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
evansiroky: This comment is not blocking and I'm just saying it'd be a nice-to-have. However, the approach wouldn't be to mock authorization bearers. If you notice, you're already making admin-only calls to create and delete labels and you're not sending any kind of mocked bearer for that. The reason this works is because of these lines of code here:

https://github.com/ibi-group/datatools-server/blob/00b5155694ef3e36d7f480351dbf4ed9ea131564/src/main/java/com/conveyal/datatools/manager/auth/Auth0Connection.java#L55-L60

So, to test out a non-admin user, perhaps the approach would be to add some kind of query parameter instructing the `Auth0Connection#checkUser` method to create a non-admin user profile for that specific request.

evansiroky (changes requested):
Just one more blocking comment: https://github.com/ibi-group/datatools-server/pull/403#discussion_r682856077.

Also see response in non-blocking comment here: https://github.com/ibi-group/datatools-server/pull/403#discussion_r682862024.

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
miles-grant-ibigroup: Good catch thank you! Slept through that one. Fixed in 4e1ce1f

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
miles-grant-ibigroup: Will add this to the technical backlog document

miles-grant-ibigroup:
Good catch and good suggestion thank you!

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
landonreed: My #410 PR has new functionality to handle testing non-admin users, which I think might be satisfactory long term.

landonreed (approved):

PR #426, opened 2021-09-29
Always trigger Pelias update job

miles-grant-ibigroup:
### Checklist

### Description

The Pelias update job wasn't firing correctly in certain cases where the server was already deployed. The Pelias update job trigger is in this PR is moved to immediately after the build, so that it only fires if the build is successful.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
@@ -319,6 +319,16 @@ public class DeployJob extends MonitorableJob {
             }
         }
 
+        // Now that the build was successful, update Pelias
+        if (deployment.peliasUpdate) {
+            // Get log upload URI from deploy job
+            AmazonS3URI logUploadS3URI = getS3FolderURI();
+
+            // Execute the pelias update job and keep track of it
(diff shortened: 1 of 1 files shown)


br648 (approved):
LGTM!

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
```diff
@@ -315,7 +315,16 @@ public void jobLogic () {
                 } catch (Exception e) {
                     status.fail(String.format("Error uploading/copying deployment bundle to s3://%s", s3Bucket), e);
                 }
-
+            }
+            
+            // Now that the build was successful, update Pelias
```
binh-dam-ibigroup, line 320: Do you mean the deployment?

binh-dam-ibigroup (approved):

src/main/java/com/conveyal/datatools/manager/jobs/DeployJob.java
miles-grant-ibigroup: It used to be deployment, but was changed to build so that the job is always fired no matter the type of deployment (since each deployment type called `return`)

PR #451, opened 2022-02-07
Update `gtfs-lib`

miles-grant-ibigroup:
### Checklist

### Description

This PR updates `gtfs-lib`. Part of that release includes https://github.com/conveyal/gtfs-lib/pull/275 which validates parent stations. As part of this, the new validator found a bug in the merge code where the stop_id is only sometimes changed, although it's always updated. We've solved this by changing the logic a little bit, but are unsure if this has unforeseen implications (although all the tests pass). @br648 as author of the code could you possibly double check that everything is still as expected?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,7 +270,7 @@
         <dependency>
             <groupId>com.github.conveyal</groupId>
             <artifactId>gtfs-lib</artifactId>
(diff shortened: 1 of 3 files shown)


br648 (approved):
I've made a couple of updates, but all good.

PR #461, opened 2022-05-05
Add /updatetrips endpoint for updating multiple trips in one request.

philip-cline:
### Checklist

### Description

This PR adds an endpoint to save multiple trips with one request, effectively moving the loop over edited timetable trips from the frontend to the backend. The new endpoint also changes saving behaviour by first deleting all of the referenced trips, and then re-saving new ones. This is done to allow trip ID swapping in one request which was previously causing errors, and which should be allowable user behaviour. The bug can be seen here: https://drive.google.com/file/d/1ho01C1DaF0apxMWDbJuRD195OeEYMgTu/view?usp=sharing

Looking for feedback on ways to improve performance, especially regarding the use of autocommits in the database connection. If you'd like, I have a front end branch which uses the new endpoint setup for testing purposes here: https://github.com/ibi-group/datatools-ui/tree/test-new-multitrip-endpoint

### DIFF
--- a/src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
+++ b/src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
@@ -119,8 +119,9 @@ public abstract class EditorController<T extends Entity> {
         // Delete entity request
         delete(ROOT_ROUTE + ID_PARAM, this::deleteOne, json::write);
 
-        // Handle special multiple delete method for trip endpoint
+        // Handle special multiple delete method and multiple update method for trip endpoint
         if ("trip".equals(classToLowercase)) {
+            post(ROOT_ROUTE + "/updatetrips", this::updateMultipleTrips, json::write);
             delete(ROOT_ROUTE, this::deleteMultipleTrips, json::write);
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
...
+            ArrayList<String> returnedJsonStrings = new ArrayList<>();
+
+            if (jsonNode == null) {
+                logMessageAndHalt(req, 400, "JSON body must be provided with trip update request.");
+            }
+            for (final JsonNode trip: trips) {
+                int id = trip.get("id").intValue();
+                // Id == -2 is the front end indication that the trip is being created.
+                if (id != -2){
+                    tableWriter.delete(id, false);
```
binh-dam-ibigroup, line 160: Could you see if you could use `tableWriter.deleteWhere(...)` to create a single delete transaction?

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
...
+                }
+            }
+            for (final JsonNode thisTrip: trips) {
+                String jsonBody = thisTrip.toString();
+                JsonNode lastTrip = trips.get(trips.size()-1);
```
binh-dam-ibigroup, line 165: Move this var to outside the loop.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
         }
     }
 
+    /**
+     * Endpoint to update multiple trips in a single transaction.
+     * Creation of this endpoint was prompted by the need to allow trip ID "flipping",
+     * where trip IDs may swap values in one request. Previously this case would cause an error
+     * since trips were saved one by one, without knowledge that a target trip ID had been "flipped".
+     */
+    private String updateMultipleTrips(Request req, Response res) {
```
binh-dam-ibigroup, line 142: Please add a unit test for this method.

binh-dam-ibigroup (changes requested):
Have to play a bit more, but it seems to be headed in the right direction. Please see adding a unit test.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
binh-dam-ibigroup: If you do a single transaction, `autocommit = false` shouldn't be necessary.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
...
+                // Id == -2 is the front end indication that the trip is being created.
+                if (id != -2){
+                    tableWriter.delete(id, false);
+                }
+            }
+            for (final JsonNode thisTrip: trips) {
+                String jsonBody = thisTrip.toString();
+                JsonNode lastTrip = trips.get(trips.size()-1);
+                Boolean isLastTrip = lastTrip.equals(thisTrip);
+                // Autocommit: true for last trip
```
binh-dam-ibigroup, line 167: Does committing the last trip also commit all the inserts before?

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
...
+                returnedJsonStrings.add(tableWriter.create(jsonBody, isLastTrip));
+            }
+            return "[" + String.join(",", returnedJsonStrings) + "]";
+        }
+        catch (Exception e) {
+            e.printStackTrace();
+            logMessageAndHalt(req, 500, "Could not save trips", e);
+        }
+        return null;
```
binh-dam-ibigroup, line 176: We should return a 200-response if success.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -132,6 +133,49 @@ private void registerRoutes() {
...
+                if (id != -2){
+                    tableWriter.delete(id, false);
+                }
+            }
+            for (final JsonNode thisTrip: trips) {
+                String jsonBody = thisTrip.toString();
+                JsonNode lastTrip = trips.get(trips.size()-1);
+                Boolean isLastTrip = lastTrip.equals(thisTrip);
+                // Autocommit: true for last trip
+                returnedJsonStrings.add(tableWriter.create(jsonBody, isLastTrip));
```
binh-dam-ibigroup, line 168: If trying to "undo" a trip_id change in a row, an error occurs after the following steps (see animation below).

Screen Recording 2022-05-06 at 1 25 05 PM

binh-dam-ibigroup (changes requested):
See the comment about editing back a trip_id.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
philip-cline: I did take a look at that method, but it's set up to receive a key field and value in the form `key = value`. We would need to refactor it to take `key IN ( ... list of ids we would pass ... )`.

PR #471, opened 2022-06-13
Update the shapefile export status bar with each feed in a deploy…

philip-cline:
…ment export

### Checklist

### Description

Updates the status bar w/ each feed GisExportJob that finishes.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/DeploymentGisExportJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/DeploymentGisExportJob.java
@@ -32,7 +32,7 @@ public class DeploymentGisExportJob extends GisExportJob {
             int feedVersionsSize = feedVersions.size();
             int currentFeedVersion = 1;
             for (SummarizedFeedVersion feedVersion: feedVersions) {
-                int percentComplete = (int) (currentFeedVersion * 100.0f) / feedVersionsSize;
+                int percentComplete = (currentFeedVersion * 100) / feedVersionsSize;
 
                 //TODO: update progress bar after each feedVersion?
                 Path outputPath = Paths.get(outDir.getPath() + "/" + feedVersion.feedSource.name);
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/manager/jobs/DeploymentGisExportJob.java
```diff
@@ -29,15 +29,21 @@ public DeploymentGisExportJob(GisExportJob.ExportType exportType, Deployment dep
         try {
             File outDir = setupGisExport();
 
+            int feedVersionsSize = feedVersions.size();
+            int currentFeedVersion = 1;
             for (SummarizedFeedVersion feedVersion: feedVersions) {
+                int percentComplete = (int) (currentFeedVersion * 100.0f) / feedVersionsSize;
```
miles-grant-ibigroup: Are there any java math util we can use to avoid this cast?

miles-grant-ibigroup (approved):
Looks and works great! Part of me thinks there might be a way to avoid the dreaded `currentFeedVersion++` but nothing off the top of my head...

src/main/java/com/conveyal/datatools/manager/jobs/DeploymentGisExportJob.java
philip-cline: How you like these apples: [34cd337](https://github.com/ibi-group/datatools-server/pull/471/commits/34cd337257dbab288e39a39a044f2f41cca2b871)

miles-grant-ibigroup: Delicious!

philip-cline:
For including an index in the forEach loop, my understanding was that this couldn't be done without creating an index manually ([ref](https://stackoverflow.com/questions/477550/is-there-a-way-to-access-an-iteration-counter-in-javas-for-each-loop)) but perhaps we could switch to an iterator if it's worth it.

binh-dam-ibigroup (approved):
LGTM

binh-dam-ibigroup:
> For including an index in the forEach loop, my understanding was that this couldn't be done without creating an index manually ([ref](https://stackoverflow.com/questions/477550/is-there-a-way-to-access-an-iteration-counter-in-javas-for-each-loop)) but perhaps we could switch to an iterator if it's worth it.

The code is fine as written.

PR #506, opened 2023-01-24
Update to Feed Source deployed feed version

br648:
### Checklist

### Description

Update to how a feed source's feed version is obtained. This now takes into consideration a pinned feed version.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
@@ -9,7 +9,6 @@ import com.conveyal.datatools.common.status.MonitorableJob;
 import com.conveyal.datatools.common.utils.Scheduler;
 import com.conveyal.datatools.common.utils.aws.CheckedAWSException;
 import com.conveyal.datatools.common.utils.aws.S3Utils;
-import com.conveyal.datatools.editor.utils.JacksonSerializers;
 import com.conveyal.datatools.manager.DataManager;
 import com.conveyal.datatools.manager.jobs.CreateFeedVersionFromSnapshotJob;
 import com.conveyal.datatools.manager.jobs.FetchSingleFeedJob;
@@ -27,8 +26,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonView;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.model.Sorts;
 import org.bson.codecs.pojo.annotations.BsonIgnore;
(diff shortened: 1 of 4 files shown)


miles-grant-ibigroup:
I will be second reviewer on this

philip-cline (approved):
This looks good for me, I'm seeing the same behaviour as before the update.

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -255,20 +276,10 @@ public void createFeedSourceWithLabels() throws IOException {
...
-        DeployJob deployJob = new DeployJob(
-            "Deploying " + deploymentDeployed.name,
-            deploymentDeployed,
-            user,
-            new OtpServer(),
-            "test-deploy",
-            DeployJob.DeployType.USE_PRELOADED_BUNDLE,
-            true
-        );
-        deployJob.run();
```
miles-grant-ibigroup, line 271: why was this removed?

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -479,77 +459,101 @@ public String latestVersionId() {
...
-    private FeedVersion getDeployedFeedVersionViaExpensiveQuery() {
+    private FeedVersion getDeployedFeedVersion() {
         if (deployedFeedVersion == null) {
-            Collection<Deployment> deployments = Persistence.deployments.getFiltered(eq("projectId", this.projectId), Sorts.descending("lastUpdated"));
-            Collection<FeedVersion> feedVersions = Persistence.feedVersions.getFiltered(eq("feedSourceId", this.id), Sorts.descending("updated"));
-            if (deployments.isEmpty() || feedVersions.isEmpty()) {
-                return null;
+            Project project = Persistence.projects.getById(this.projectId);
+            if (project.pinnedDeploymentId != null) {
+                Deployment deployment = Persistence.deployments.getById(project.pinnedDeploymentId);
```
miles-grant-ibigroup: I believe this line is causing the slowdown. Would it be possible to cache the deployments in memory on startup and try searching that cache before resorting to this?

miles-grant-ibigroup (changes requested):
Works really well and behavior is correct! However, the slowdown is significant, and I think that should be addressed before we merge

philip-cline:
Also thinking out loud here, but could we potentially keep the old code but only update the mongo property when a pinned deployment is deployed? Or we still have a separate issue with other information being updated there?

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
br648: It's not needed because I also removed the code related to this task from the DeployJob. Running this job no longer has and effect on defining the deployed feed version.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
br648: I have made some further updates to reduce objects being returned and in the case of deployment objects limiting the fields returned. I'm hoping this is enough to avoid caching as managing that will be quite onerous.

src/main/java/com/conveyal/datatools/manager/persistence/TypedPersistence.java
```diff
@@ -176,11 +188,20 @@ public List<T> getFiltered (Bson filter) {
...
-        }
+        return (sortBy != null)
+        ? mongoCollection.find(filter).sort(sortBy).into(new ArrayList<>())
+        : mongoCollection.find(filter).into(new ArrayList<>());
+    }
+
+    /**
+     * Get all objects populating only the included fields satisfying the supplied Mongo filter and sort by.
+     */
+    public List<T> getFilteredLimitedFields(Bson filter, Bson sortBy, String... includedFields) {
```
miles-grant-ibigroup, line 199: It's better than before but it's still quite slow... 

Is there any way to make the request only once on startup and then hold it? that should be less extreme than fully caching the results. we could fallback to making this request if we don't find the data in that initial request data.

I tried fiddling with mongo settings but couldn't get performance better enough

miles-grant-ibigroup (approved):
Going to approve this so this is fixed, but think we should still attempt the initial request method if possible! If you want to discuss more outside of GitHub let me know

miles-grant-ibigroup:
Let's merge this, then move forward with our approach of moving the expensive fields to a new endpoint

PR #508, opened 2023-02-02
Mtc optimize feedupdater

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected => This is a patch on the MTC deployment branch. A similar PR will eventually be issued to the `dev` branch.
- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing => Semantic release does not pick up the correct Node version.

### Description

This PR improves performance and memory usage of the `FeedUpdater` feature:
- the query that populates/updates `versionsToMarkAsProcessed` uses less memory by only requesting the `id` field (and using `FeedVersionSummaries` instead of `FeedVersion`) (see comparison charts below). A bug in the query is also fixed.
- A single mongo query is performed, instead of one per agency, to get a map of all feed sources by agency (what MTC calls "feed id").
- A single mongo query is performed, instead of one per agency, to get a map of the latest feed version sent to publishing for each feed source.

Screenshots of memory profiles (left: current, right: with PR changes):

image

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
@@ -12,6 +12,7 @@ import com.conveyal.datatools.manager.models.FeedVersion;
 import com.conveyal.datatools.manager.persistence.FeedStore;
 import com.conveyal.datatools.manager.persistence.Persistence;
 import com.conveyal.datatools.manager.utils.HashUtils;
+import com.google.common.collect.Lists;
 import com.google.common.io.ByteStreams;
 import org.bson.conversions.Bson;
 import org.slf4j.Logger;
@@ -30,12 +31,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static com.conveyal.datatools.manager.extensions.mtc.MtcFeedResource.AGENCY_ID_FIELDNAME;
 import static com.conveyal.datatools.common.utils.Scheduler.schedulerService;
 import static com.mongodb.client.model.Filters.and;
 import static com.mongodb.client.model.Filters.eq;
+import static com.mongodb.client.model.Filters.in;
 import static com.mongodb.client.model.Filters.lt;
 import static com.mongodb.client.model.Filters.ne;
 import static com.mongodb.client.model.Filters.or;
@@ -101,8 +104,29 @@ public class FeedUpdater {
         return new FeedUpdater(completedFeedRetriever);
     }
 
+    private static String getFeedId(S3ObjectSummary objSummary) {
+        String[] parts = objSummary.getKey().split("/");
+        return parts.length > 1 ? parts[1].replace(".zip", "") : "";
+    }
+
+    public static List<String> getFeedIds(List<S3ObjectSummary> s3objects) {
+        return s3objects.stream()
+            .map(FeedUpdater::getFeedId)
+            .filter(id -> id.length() != 0)
+            .collect(Collectors.toList());
+    }
+
+    /**
+     * @return The property objects for the specified feed id.
+     */
+    public static List<ExternalFeedSourceProperty> getPropertiesForFeedId(List<ExternalFeedSourceProperty> properties, String value) {
+        return properties.stream().filter(prop -> value.equals(prop.value)).collect(Collectors.toList());
+    }
+
     private class UpdateFeedsTask implements Runnable {
         public void run() {
+            LOG.info("Starting UpdateFeedsTask");
+            long startMillis = System.currentTimeMillis();
             Map<String, String> updatedTags;
             try {
                 LOG.debug("Checking MTC feeds for newly processed versions");
@@ -113,6 +137,8 @@ public class FeedUpdater {
             } catch (Exception e) {
                 LOG.error("Error updating feeds {}", e);
             }
+
+            LOG.info("Completed UpdateFeedsTask in {} ms", System.currentTimeMillis() - startMillis);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
```diff
@@ -0,0 +1,235 @@
...
+            prop1,
+            prop2,
+            prop3
+        );
+
+        Map<String, String> feedIdsToFeedSourceIds = FeedUpdater.mapFeedIdsToFeedSourceIds(properties);
+
+        assertEquals(FEED_SOURCE_1_ID, feedIdsToFeedSourceIds.get(FIRST_FEED));
+        assertEquals("other-feed-source-id", feedIdsToFeedSourceIds.get(OTHER_FEED));
+        assertNull(feedIdsToFeedSourceIds.get("unknownFeed"));
```
br648: What is the purpose of this test? The outcome will always be null because `feedIdsToFeedSourceIds` will never contain "unknownFeed".

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -101,12 +111,54 @@ public static FeedUpdater createForTest(CompletedFeedRetriever completedFeedRetr
...
+
+    public static Map<String, FeedSource> mapFeedIdsToFeedSources(
+        Map<String, String> feedIdsToFeedSourceIds,
+        List<FeedSource> feedSources
+    ) {
+        Map<String, FeedSource> feedSourceIdToFeedSource = feedSources.stream()
+            .collect(Collectors.toMap(src -> src.id, Function.identity()));
+
+        return feedIdsToFeedSourceIds.entrySet().stream()
+            .collect(Collectors.toMap(e -> e.getKey(), e -> feedSourceIdToFeedSource.get(e.getValue())));
```
br648: SonarLint: `e -> e.getKey()` can be replaced with `Map.Entry::getKey`. Although it seems clearer to me to leave as is!

src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
```diff
@@ -0,0 +1,235 @@
...
+        List<String> expectedIds = Lists.newArrayList(
+            FIRST_FEED,
+            OTHER_FEED
+        );
+        assertEquals(expectedIds, FeedUpdater.getFeedIds(s3objects));
+    }
+
+    @Test
+    void shouldMapFeedIdsToFeedSourceIds() {
+        ExternalFeedSourceProperty prop1 = new ExternalFeedSourceProperty();
```
br648: Consider creating a helper method to create `ExternalFeedSourceProperty` instances.

src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
```diff
@@ -0,0 +1,235 @@
...
+
+        Map<String, FeedSource> feedIdToFeedSource = FeedUpdater.mapFeedIdsToFeedSources(feedIdsToFeedSourceIds, feedSources);
+        assertEquals(fs1, feedIdToFeedSource.get(FIRST_FEED));
+        assertEquals(fs2, feedIdToFeedSource.get(OTHER_FEED));
+    }
+
+    @Test
+    void shouldGetLatestVersionsSentToPublisher() {
+        // Create two feed sources, and two feed versions in each that have different sentToExternalPublisher values.
+        FeedSource source1 = new FeedSource(FEED_SOURCE_1_ID);
```
br648: Consider creating helper methods to create `FeedSource` and `FeedVersion` instances.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
```diff
@@ -264,18 +322,75 @@ private void updatePublishedFeedVersion(String feedId, FeedVersion publishedVers
...
+            "$replaceRoot": {
+                "newRoot": "$items"
+            }
+        }
+        ])
+        */
+
+        List<String> feedSourceIds = feedSources.stream().map(fs -> fs.id).collect(Collectors.toList());
+
+        List<Bson> stages = Lists.newArrayList(
```
br648, line 359: Is there potential for this to fail silently if there is a change to the underlying fields? Could this be guarded against perhaps by checking for the existence of all required fields first?

br648 (commented):
Just a few comments for consideration. Interesting low level Mongo queries.

src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
binh-dam-ibigroup: That assertion would have covered the uses of the `resourceType` field, but it doesn't seem to be used for anything except the resources tabs jn the feed settings. I removed that assertion.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-server/pull/508/commits/3bb3b32530a87f525317d0b46cc21e71ed11067a.

src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
binh-dam-ibigroup: Yes, I was on the fence but I've refactored in https://github.com/ibi-group/datatools-server/pull/508/commits/3bb3b32530a87f525317d0b46cc21e71ed11067a.

src/test/java/com/conveyal/datatools/manager/jobs/FeedUpdaterTest.java
binh-dam-ibigroup: Refactored in https://github.com/ibi-group/datatools-server/pull/508/commits/3bb3b32530a87f525317d0b46cc21e71ed11067a.

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
binh-dam-ibigroup: The check for the only required field `sentToExternalPublisher` besides `_id` is already being done in the `match` clause.

br648: If I change `feedSourceId` or `latestSentToExternalPublisher` to "TEST", Mongo does not throw an SQL exception. Perhaps the test `shouldGetLatestVersionsSentToPublisher` failing is enough to cover this if fields are changed in the future? I understand that this is an edge case, but an RDMS would throw an exception for missing/unknown field.

br648 (commented):
Just need your thoughts on this: https://github.com/ibi-group/datatools-server/pull/508#discussion_r1098502741

src/main/java/com/conveyal/datatools/manager/jobs/FeedUpdater.java
binh-dam-ibigroup: Correct, this has to do with how Mongo fundamentally works. If the field is renamed for some reason, the publishing feature will silently fail, but also the test `shouldGetLatestVersionsSentToPublisher` that ties the Java field with the Mongo field will fail too as you mention, so we should be covered.

br648 (approved):
All good. We should look to use Mongo queries (as used here) more often.

philip-cline (approved):
LGTM

PR #512, opened 2023-02-16
Add MobilityData GTFS Validator to validation step

miles-grant-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR integrates the [MobilityData GTFS Validator](https://github.com/MobilityData/gtfs-validator) into datatools, and adds a MobilityData validation to the GTFS validation step. Validation results are saved to a unique temporary folder, then loaded and inserted into Mongo. Failures are handled via the existing validation failure infrastructure.

#### Room for Improvement
- [x] I'd like to perhaps see the MobilityData validation happen in a second thread, possibly in parallel to the existing validation. However, I'm not sure how to do this in the cleanest way, and the new validation isn't that slow, especially compared to the existing one.

#### Important Notices
The MobilityData validator requires a much newer version of the Java spec. I've updated by JDK to 18, and language level to 11!

**Tests are failing due to an unrelated shape file export issue**

PR #513, opened 2023-02-17
Threaded validation

br648:
### Checklist

- [ ] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [ ] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [ ] The description lists all applicable issues this PR seeks to resolve
- [ ] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing

### Description

Working on top of https://github.com/ibi-group/datatools-server/pull/512/files original and mobility data validation are now in their own separate threads.


br648:
@binh-dam-ibigroup, Miles is to carry out some testing, but on the whole is happy to merge this into his working PR. If you don't mind, can you review please just in case I've missed something?

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -382,22 +414,6 @@ public void validate(MonitorableJob.Status status) {
 
             // This will persist the document to Mongo
             this.mobilityDataResult = Document.parse(json);
```
miles-grant-ibigroup, line 418: The "regular" parsing step was previously persisting the `this` to mongo. If the mobility data parser finishes first, this still happens. If it finishes second, this fails to happen and the data isn't persisted! I think the solution is to add a mongo persistence line right here

miles-grant-ibigroup (changes requested):
Seems to be working nice and fast, only I noticed to resolve is a race condition that sometimes prevents results from being persisted to mongo

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
br648: @miles-grant-ibigroup We need to be careful here because we don't want to create a new Mongo object if the other validator is expecting to do that. See ValidateFeedJob -> jobFinished(). Both validators need to be aware of the Mongo object state, so that either can create or update. Let me take a look. Good spot btw.

br648: @miles-grant-ibigroup This should now be addressed. Unit tests look good. Lemme know how you get on.

src/main/java/com/conveyal/datatools/manager/jobs/ValidateFeedJob.java
```diff
@@ -44,13 +44,7 @@ public void jobFinished () {
...
                 // the version won't get loaded into MongoDB (even though it exists in postgres).
                 feedVersion.storeUser(owner);
-                if (isNewVersion) {
-                    int count = feedVersion.parentFeedSource().feedVersionCount();
-                    feedVersion.version = count + 1;
-                    Persistence.feedVersions.create(feedVersion);
-                } else {
-                    Persistence.feedVersions.replace(feedVersion.id, feedVersion);
-                }
+                feedVersion.persistFeedVersionAfterValidation(isNewVersion);
```
binh-dam-ibigroup, line 47: As written, it looks like the new version is shown in the UI (on the `mobility-data-validator` branch) after the first validator completes. Is that the desired effect?

miles-grant-ibigroup: I think that's ok as long as we refresh the UI when the mobility data results come in, which is something we can do on the client.

miles-grant-ibigroup (approved):
Things are working well for me, I think this is good to merge back into the main mobility data branch thanks for the big improvement!

miles-grant-ibigroup:
I've noticed that on some large feeds the jobs are running sequentially?
Screenshot 2023-03-01 at 11 50 35 AM

br648:
This is to be expected. Once both threads have started we are then effectively passing control to the JVM which then decides how much CPU each thread will get. If the first gets 100% CPU, the second will have to wait.

binh-dam-ibigroup (approved):
Good deal!

PR #514, opened 2023-02-17
Return otpServers with /project endpoint to allow speedup

philip-cline:
### Checklist

### Description

Recent front end PRs (https://github.com/ibi-group/datatools-ui/pull/895) have tried to reduce the amount of data that is unnecessarily requested by the front end. This endpoint created a conflict bc the /project and project:id differed in their return structure (specifically in returning otpServers). 

Per the comment in the backend, I'm worried about performance issues with this change: 
image

TODO: perhaps the ProjectWithOtpServers should not be a mixin anymore?

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/ProjectController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/ProjectController.java
@@ -343,16 +343,17 @@ public class ProjectController {
      */
     public static void register (String apiPrefix) {
         // Construct JSON managers which help serialize the response. Slim JSON is the generic JSON view. Full JSON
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
LGTM

br648:
@philip-cline following on from your comment, have you noticed any performance issues with this change and is this related to your TODO? I.e. if there is a performance issue, should this update be rolled back?

philip-cline:
Let me do some testing on a large project deployment and see. The TODO is separate from the performance issue, that is more just for if this fix is permanent, then it doesn't really make sense to keep the field as a mixin.

philip-cline:
Yeah this seems to slow things down significantly for a big project. I'll do some work to maybe reduce the amount of data we need to request.

PR #516, opened 2023-02-28
Deployed feed version summaries

br648:
### Checklist

### Description

A new end point has been added to the feed source controller to return the deployed feed version for a given feed source. If the project has a pinned deployment and this deployment has a feed version this is used. If not, the latest feed version for a feed source is used.


br648:
@philip-cline I've reverted this back to using the previous approach of getting the deployed feed version with the feed source, so will work with the existing UI functionality. I'm still slightly concerned by how long this might take with 100+ feed sources. Please let me know how it performs!

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -459,106 +459,59 @@ public String latestVersionId() {
...
-                }
-            }
-        }
-        return deployedFeedVersion;
-    }
-
-    /**
-     * Get the latest deployed feed version for this feed source, if available.
+     * If a project has a "pinned" deployment, that deployment's feed versions take precedence over the latest
+     * deployment's feed versions for this feed source. In this case, return the feed version from the pinned deployment.
```
philip-cline: I would maybe amend this comment to say that a pinned deployment supercedes all other deployments, such that if there is a pinned deployment and a feed is not included in it, there is no deployedFeedVersion for that feed.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -459,106 +459,59 @@ public String latestVersionId() {
...
-            Sorts.descending("updated")
-        );
+    public FeedVersionDeployed retrieveDeployedFeedVersion() {
+        if (deployedFeedVersion != null) {
+            return deployedFeedVersion;
+        }
+        deployedFeedVersion = FeedVersionDeployed.getFeedVersionFromPinnedDeployment(projectId, id);
+        return (deployedFeedVersion != null)
+            ? deployedFeedVersion
+            : FeedVersionDeployed.getFeedVersionFromLatestDeployment(projectId, id);
```
philip-cline, line 522: Are we returning nothing if there is no pinned deployment?

philip-cline (approved):
This is looking good to me. It has sped things up for a large deployment to around 5s, which maybe isn't great but it's a start at least.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -459,108 +459,73 @@ public String latestVersionId() {
...
+     *
+     * If a project does not have a "pinned" deployment, return the latest deployment's feed versions for this feed
+     * source, if available.
+     */
+    public FeedVersionDeployed retrieveDeployedFeedVersion() {
+        if (deployedFeedVersionDefined) {
+            return deployedFeedVersion;
+        }
+        Project project = Persistence.projects.getById(projectId);
+        if (project.pinnedDeploymentId != null && !project.pinnedDeploymentId.isEmpty()) {
```
miles-grant-ibigroup: I suspect this call is what's slowing down things significantly. What do we think about making this call optional? @philip-cline

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
```diff
@@ -459,108 +459,73 @@ public String latestVersionId() {
...
+    @BsonIgnore
+    private FeedVersionDeployed deployedFeedVersion;
+
+    /**
+     * This value is set to true once an attempt has been made to get the deployed feed version. This prevents subsequent
+     * attempts where the deployed feed version is not available.
+     */
+    @JsonIgnore
+    @BsonIgnore
+    private boolean deployedFeedVersionDefined;
```
miles-grant-ibigroup, line 477: Why do we do this instead of a null check on deployedFeedVersion?

src/main/java/com/conveyal/datatools/manager/models/FeedVersionDeployed.java
```diff
@@ -0,0 +1,239 @@
...
+            {
+               $limit: 1
+            }
+            ])
+         */
+        List<Bson> stages = Lists.newArrayList(
+            match(
+                in("_id", projectId)
+            ),
+            lookup("Deployment", "pinnedDeploymentId", "_id", "deployment"),
```
miles-grant-ibigroup, line 111: Is there no way to check for the pinned deployment existing as part of this call?

miles-grant-ibigroup (approved):
Working great! Just a few possibilities for speed improvements possibly, but I don't have any concrete suggestions so will approve while putting these ideas out there. Let me know what you think!

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
br648: This is to prevent calls to `retrieveDeployedFeedVersion()` where there are no deployed feed versions. Using null resulted in the following scenerio: Call to `getDeployedFeedVersionId()` returned null, followed by a call to `getDeployedFeedVersionStartDate()` returning null and a call to `getDeployedFeedVersionEndDate()` also returning null.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
br648: I would liked to have included this in the Mongo query, but hit my limitations with Mongo. If this is an issue I can look at this again and attempt to include?

src/main/java/com/conveyal/datatools/manager/models/FeedVersionDeployed.java
br648: Short answer, probably. See my response to question above, but I would need to spend more time on this.

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: I agree this can be a separate PR!

src/main/java/com/conveyal/datatools/manager/models/FeedSource.java
miles-grant-ibigroup: I think this may just be issues with my java understanding how is this field better than those 3 methods returning null? Is it to avoid mongo queries?

src/main/java/com/conveyal/datatools/manager/models/FeedVersionDeployed.java
miles-grant-ibigroup: Let's make it a separate PR!

br648:
Going to merge this to get the correct functionality to prod. Then look to include the project DB call into the two Mongo queries to further improve performance in a new PR.

PR #525, opened 2023-03-24
fix(MobilityDataValidation): slow on large gtfs files

### LINKED ISSUE #524: MobilityData Validator CSV Parsing Issue with large GTFS files
## Observed behavior

MobilityData Validation Result can run into CSV parsing issues when validating the MBTA's GTFS feed. Running the same feed with the MobilityData Validator locally does not produce the same result.  

## Expected behavior

Validator should not produce the parsing issue. 

Validate MBTA's latest GTFS feed. Observe the following errors: 
image

None 

Latest validator branch, dev server.

miles-grant-ibigroup:
### Checklist

### Description

The MobilityData validator takes a while between finishing and saving results. There is no way to check when this process completes using their api. The solution is to do a wait, currently 5 seconds. However, on large feeds this is not enough. The solution is to adjust the wait depending on file size.

Closes #524 

Looking to try this fix on as wide a variety of systems as possible to make sure the values are correct

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
@@ -393,8 +393,8 @@ public class FeedVersion extends Model implements Serializable {
             status.update("MobilityData Analysis...", 11);
 
             // Wait for the file to be entirely copied into the directory.
-            // TODO: base this on the file being completely saved rather than a fixed amount of time.
-            Thread.sleep(5000);
+            // 5 seconds + ~1 second per 10mb
+            Thread.sleep(5000 + (this.fileSize / 10000));
             File gtfsZip = this.retrieveGtfsFile();
             // Namespace based folders avoid clash for validation being run on multiple versions of a feed.
             // TODO: do we know that there will always be a namespace?
--- a/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
@@ -393,7 +393,7 @@ public class FeedVersion extends Model implements Serializable {
             status.update("MobilityData Analysis...", 11);
 
             // Wait for the file to be entirely copied into the directory.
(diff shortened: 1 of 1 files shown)


philip-cline (approved):

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -393,8 +393,8 @@ public void validateMobility(MonitorableJob.Status status) {
             status.update("MobilityData Analysis...", 11);
 
             // Wait for the file to be entirely copied into the directory.
-            // TODO: base this on the file being completely saved rather than a fixed amount of time.
-            Thread.sleep(5000);
+            // 5 seconds + ~1 second per 10mb
+            Thread.sleep(5000 + (this.fileSize / 10000));
```
binh-dam-ibigroup, line 397: What unit is `this.fileSize`? According to the code, it is set with `[newGtfs]File.length()` which is in bytes.

miles-grant-ibigroup: I'm not sure, but this does fix things

binh-dam-ibigroup: @miles-grant-ibigroup Can you fix the comment so it corresponds to the code?

miles-grant-ibigroup: Did I get the decimal math right in 0ebf2227078653ede1d54a77676b22bb4a89a726?

binh-dam-ibigroup: I'm sorry, the initial comment was correct, juggling with the units was confusing (10,000,000 / 10,000 = 1000).

binh-dam-ibigroup (approved):
It was indeed 1 sec / 10 MB...

miles-grant-ibigroup:
Not sure why it says closed instead of merged...

PR #529, opened 2023-04-18
Feed Source Summary

br648:
### Checklist

### Description

Update to provide a new end point which will return all feed source summaries as required by the UI.

Companion PR: https://github.com/ibi-group/datatools-ui/pull/981

See: https://ibisolutions.atlassian.net/browse/DT-315?focusedCommentId=251321


binh-dam-ibigroup:
@br648 I think the main bottleneck is that the same query is still repeated for each project, and as a result, the number of times we hit Mongo doesn't change overall. Could you see what happens if `getFeedVersionFromPinnedDeployment` and `getFeedVersionFromLatestDeployment` grabbed all projects at once?

philip-cline:
Couple properties we need in the `FeedSourceSummary` for the front end: 
- labelIds 
- latestValidation to show this portion of the FeedSource table: 
image
For this section, we need to query the latest feedVersion - is that something that should be wrapped up in the project query for efficiency?

br648:
@philip-cline I've made an update to include feed source label ids. 

The latest feed version should already be available as part of the information returned by feedsourceSummaries. To keep the payload to a minimum, only the required values from the latest feed version are returned. See `FeedSourceControllerTest.canRetrieveDeployedFeedVersionFromLatestDeployment()` and `FeedSourceControllerTest.canRetrieveDeployedFeedVersionFromPinnedDeployment()`. Let me know if additional values are required.

philip-cline:
@br648 Would it be easy to include `projectId` in the summary data? Any feedsource update after getting the feedsource summaries requires the project ID to make an update to the feedsource. I could update all the front end methods to include this, but it seems easier to just include it in the data in the first place

br648:
> @br648 Would it be easy to include `projectId` in the summary data? Any feedsource update after getting the feedsource summaries requires the project ID to make an update to the feedsource. I could update all the front end methods to include this, but it seems easier to just include it in the data in the first place

@philip-cline I have added the project id to the feed source summary.

philip-cline:
It seems like it will be easier on the front end to integrate these changes if we maintain some of the same structure as previous. Could we use the following structure: 
```
{
  "projectId": "...",
  "id": "...",
  "name": "...",
  "deployable": true,
  "isPublic": false,
  "lastUpdated": "...",
  "labelIds": [],
  "deployedFeedVersionId": "...",
  "deployedFeedVersionStartDate": "...",
  "deployedFeedVersionEndDate": "...",
  "deployedFeedVersionIssues": X,
  latestValidation: {
  		"feedVersionId": "...",
  		"feedVersionStartDate": "...",
  		"feedVersionStartDate": "...",
  		"feedVersionIssues": "..."
  		
  }
}
```

br648:
@philip-cline Structure updated as requested.

philip-cline:
Thanks for the update @br648! I just made another small commit changing the field names so that they require less work on the front end

philip-cline (approved):
This is looking good to me! I have been able to create a front end PR to test it out and works well!

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+     * Set the appropriate feed version. For consistency, if no error count is available set the related number of
+     * issues to null.
+     */
+    public void setFeedVersion(FeedVersionSummary feedVersionSummary, boolean isDeployed) {
+        if (feedVersionSummary != null) {
+            if (isDeployed) {
+                this.deployedFeedVersionId = feedVersionSummary.id;
+                this.deployedFeedVersionStartDate = feedVersionSummary.validationResult.firstCalendarDate;
+                this.deployedFeedVersionEndDate = feedVersionSummary.validationResult.lastCalendarDate;
+                this.deployedFeedVersionIssues = (feedVersionSummary.validationResult.errorCount == -1)
```
binh-dam-ibigroup, line 92: What is the difference between a `null` issues list vs. an issues list of size zero?

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+                        from:"Deployment",
+                        localField:"pinnedDeploymentId",
+                        foreignField:"_id",
+                        as:"deployment"
+                    }
+				},
+				{
+					$unwind: "$deployment"
+				},
+				{
```
binh-dam-ibigroup: Remove excess indent.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+                    $match: {
+                        _id: "project-with-pinned-deployment"
+                    }
+                },
+                {
+                    $project: {
+                        pinnedDeploymentId: 1
+                    }
+                },
+				{
```
binh-dam-ibigroup: Remove excess indent.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+					$unwind: "$deployment"
+				},
+				{
+                    $lookup:{
+                        from:"FeedVersion",
+                        localField:"deployment.feedVersionIds",
+                        foreignField:"_id",
+                        as:"feedVersions"
+                    }
+				},
```
binh-dam-ibigroup: Remove excess indent.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+                Accumulators.first("feedVersionId", "$feedVersions._id"),
+                Accumulators.first("firstCalendarDate", "$feedVersions.validationResult.firstCalendarDate"),
+                Accumulators.first("lastCalendarDate", "$feedVersions.validationResult.lastCalendarDate"),
+                Accumulators.first("errorCount", "$feedVersions.validationResult.errorCount")
+            )
+        );
+        return extractFeedVersionSummaries(
+            "FeedSource",
+            "feedVersionId",
+            "_id",
```
binh-dam-ibigroup, line 212: Are these arguments swapped intentionally compared to the 2 other places this function is called?

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
```diff
@@ -0,0 +1,548 @@
...
+    public FeedSourceSummary() {
+    }
+
+    public FeedSourceSummary(String projectId, Document feedSourceDocument) {
+        this.projectId = projectId;
+        this.id = feedSourceDocument.getString("_id");
+        this.name = feedSourceDocument.getString("name");
+        this.deployable = feedSourceDocument.getBoolean("deployable");
+        this.isPublic = feedSourceDocument.getBoolean("isPublic");
+        this.labelIds = feedSourceDocument.getList("labelIds", String.class);
```
binh-dam-ibigroup: The UI currently assumes that `labelIds` is non-null and at least an empty list. I think this line may cause `null` to be returned in the response and break the UI.
@philip-cline What do you think?

binh-dam-ibigroup (approved):
Inclined to approve, as the loading is much faster with the UI PR, but I would make sure `labelIds` is non-null.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
philip-cline: I haven't experienced that in a project without labelIds, it seems to return an empty array

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
br648: I've updated from null to zero as the value comes from a primitive which would default to zero.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
br648: I've updated to always return an empty array if labelIds is null.

src/main/java/com/conveyal/datatools/manager/models/FeedSourceSummary.java
br648: Yep! The values to be extracted are different here because it's working with feed source documents. The feed source key in this case is the document id (_id) and not a document field (in the case of the other calls).

PR #539, opened 2023-05-17
Deletion of a stop from pattern stops

br648:
### Checklist

### Description

Added a new end point which allows a stop to be deleted along with all references in stop times and pattern stops. Once deleted the altered pattern stops are normalized to address the removed stop.

Jira: https://ibisolutions.atlassian.net/browse/DT-322

### DIFF
--- a/src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
+++ b/src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
@@ -29,8 +29,10 @@ import javax.sql.DataSource;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -41,6 +43,7 @@ import static com.conveyal.datatools.common.utils.SparkUtils.formatJSON;
 import static com.conveyal.datatools.common.utils.SparkUtils.getObjectNode;
 import static com.conveyal.datatools.common.utils.SparkUtils.logMessageAndHalt;
 import static com.conveyal.datatools.manager.controllers.api.UserController.inTestingEnvironment;
+import static org.eclipse.jetty.http.HttpStatus.OK_200;
 import static spark.Spark.delete;
 import static spark.Spark.options;
 import static spark.Spark.patch;
@@ -128,6 +131,10 @@ public abstract class EditorController<T extends Entity> {
         if ("pattern".equals(classToLowercase)) {
             put(ROOT_ROUTE + ID_PARAM + "/stop_times", this::updateStopTimesFromPatternStops, json::write);
             delete(ROOT_ROUTE + ID_PARAM + "/trips", this::deleteTripsForPattern, json::write);
+        }
+
+        if ("stop".equals(classToLowercase)) {
+            patch(ROOT_ROUTE + "/deletefrompatternstops", this::deleteStopFromPatternStops, json::write);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -258,6 +265,65 @@ private String deleteTripsForPattern(Request req, Response res) {
...
+        // NOTE: This is a string stop ID, not the integer ID that all other HTTP endpoints use.
+        String stopId = req.queryParams("stopId");
+        if (stopId == null) {
+            logMessageAndHalt(req, 400, "Must provide a valid stopId.");
+        }
+
+        try (
+            Connection connection = datasource.getConnection();
+            PreparedStatement statement = connection.prepareStatement(
+                String.format("select pattern_id, stop_sequence from %s.pattern_stops where stop_id = ?", namespace)
```
philip-cline: Would it be cleaner to format with the `stopId` here?

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -258,6 +265,65 @@ private String deleteTripsForPattern(Request req, Response res) {
...
+            )
+        ) {
+            // Get the patterns to be normalized before the related stop is deleted.
+            statement.setString(1, stopId);
+            ResultSet resultSet = statement.executeQuery();
+            Map<Integer, Integer> patternsToBeNormalized = new HashMap<>();
+            while (resultSet.next()) {
+                patternsToBeNormalized.put(
+                    resultSet.getInt("pattern_id"),
```
philip-cline: I don't think we're guaranteed that the `patternId` is an `int`: 
image

philip-cline (changes requested):
Looking really good for pattern_stops, but have a few comments and we'll need to remove the stop from stop_time entries as well.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
br648: Good spot. I should be using the id and not pattern_id (they are the same in the test data). Fixed.

philip-cline: Yeah it's tricky. that got me too with the other pattern stop PR

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
br648: Sorry I don't follow, can you elaborate please?

philip-cline: Oh I meant like `String.format("select id, stop_sequence from %s.pattern_stops where %s = %s", namespace, stopIdColumnName, stopId)` so using the String format rather than the PreparedStatement parameter

philip-cline: I'm not really sure it's cleaner, just a thought since we're not really reusing the PreparedStatement

philip-cline:
Merged dev just so we could get the pattern id updates for testing

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
br648: I happened to address this in a follow up commit, so all good.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
```diff
@@ -129,6 +132,10 @@ private void registerRoutes() {
             put(ROOT_ROUTE + ID_PARAM + "/stop_times", this::updateStopTimesFromPatternStops, json::write);
             delete(ROOT_ROUTE + ID_PARAM + "/trips", this::deleteTripsForPattern, json::write);
         }
+
+        if ("stop".equals(classToLowercase)) {
+            patch(ROOT_ROUTE + "/cascadeDeleteStop", this::cascadeDeleteStop, json::write);
```
philip-cline: Could we make the endpoint a `delete` or just a `get` endpoint? `patch` requests have some annoying requirements that make things a bit difficult on the front end

src/test/java/com/conveyal/datatools/editor/controllers/api/EditorControllerTest.java
```diff
@@ -143,6 +147,60 @@ public void canPatchStopsConditionally() throws IOException {
...
+        assertThatSqlCountQueryYieldsExpectedCount(patternStopsCountSql, 4);
+
+        String path = String.format(
+            "/api/editor/secure/stop/cascadeDeleteStop?stopId=%s&feedId=%s&sessionId=test",
+            stopId,
+            feedVersion.feedSourceId
+        );
+        String response = given()
+            .port(DataManager.PORT)
+            .patch(path)
```
philip-cline: Just marking that this would have to be updated if we change the endpoint type

philip-cline (approved):
Testing this from some rough front end code looked good. Thanks for all the changes.

src/main/java/com/conveyal/datatools/editor/controllers/api/EditorController.java
br648: @philip-cline I've changed to a delete. Note that the stop id now has to be part of the URL and not a query param:
`/api/editor/secure/stop/STOP_ID_HERE/cascadeDeleteStop?feedId=%s&sessionId=test`

src/test/java/com/conveyal/datatools/editor/controllers/api/EditorControllerTest.java
br648: Noted and updated to delete.

miles-grant-ibigroup (approved):
Working well and code looks clean thanks for these changes!

PR #555, opened 2023-08-16
Disable auth for tests

br648:
### Checklist

### Description

73 exceptions are logged as part of CI for failed authentication. Auth isn't needed for the majority of unit tests, so where the exception is thrown, auth has been disabled. It is re-enabled once the test has completed so as not to impact tests that do require auth. 

This PR (where applicable) disables auth which will remove the exceptions logged and speed up testing... a little.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -268,7 +268,7 @@
             <groupId>com.github.conveyal</groupId>
             <artifactId>gtfs-lib</artifactId>
             <!-- Latest dev build on jitpack.io -->
-            <version>f2ceb59</version>
+            <version>f2ceb59027</version>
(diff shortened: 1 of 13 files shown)


miles-grant-ibigroup (approved):
Is there no better place to put all these calls? Feels annoying to have to repeat it so many times

br648:
> Is there no better place to put all these calls? Feels annoying to have to repeat it so many times

@miles-grant-ibigroup not without a more indepth refactor. Some of the tests require auth so a global update (say in `DatatoolsTest.setUp()`) is not possible. If I was to do this, I'd have to update all the tests that require auth (to enable it), so I don't think we'd be any further forward.

binh-dam-ibigroup (approved):

PR #552, opened 2023-08-03
Shared Stops Support

miles-grant-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing
- [x] The description lists all relevant PRs included in this release _(remove this if not merging to master)_
- [x] e2e tests are all passing _(remove this if not merging to master)_

### Description

Adds support for Shared Stops feature in OTP.

This PR relies on https://github.com/conveyal/gtfs-lib/pull/391. Everything is erroring until that is updated in pom.xml. If you manually set the gtfs-lib version to `9903b3b63ad26d135b5a55cf17e595557db037bb`, this PR will work.

PR #557, opened 2023-08-24
Shared Stops Validator: Validate `feed_id`

miles-grant-ibigroup:
### Checklist

### Description

Builds on #552, only validating the shared stop if the feed_id of the feed matches the feed id of the row being validated


src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -372,7 +373,14 @@ public void validate(MonitorableJob.Status status) {
...
             } else {
                 FeedSource fs = Persistence.feedSources.getById(this.feedSourceId);
-                SharedStopsValidator ssv = new SharedStopsValidator(fs.retrieveProject());
+
+                // Get feed_id from feed version... Really awful hack!
+                JDBCFetcher feedFetcher = new JDBCFetcher("feed_info", null);
+                Object gtfsFeedId = feedFetcher.getResults(this.namespace, null, null).get(0).get("feed_id");
+
+
+                String feedId = gtfsFeedId == null ? "" : gtfsFeedId.toString();
```
br648, line 390: I think the `""` should be `null` because of this check: `if (this.feedId == null) {` in ` SharedStopsValidator`

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -372,7 +373,14 @@ public void validate(MonitorableJob.Status status) {
                 );
             } else {
                 FeedSource fs = Persistence.feedSources.getById(this.feedSourceId);
-                SharedStopsValidator ssv = new SharedStopsValidator(fs.retrieveProject());
+
+                // Get feed_id from feed version... Really awful hack!
+                JDBCFetcher feedFetcher = new JDBCFetcher("feed_info", null);
```
br648: Change to: `new JDBCFetcher("feed_info");`

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
```diff
@@ -24,24 +23,30 @@ public class SharedStopsValidator extends FeedValidator {
...
-    public SharedStopsValidator(Project project) {
+    public SharedStopsValidator(Project project, String feedId) {
         super(null, null);
         this.project = project;
+        this.feedId = feedId;
     }
 
     // This method can only be run on a SharedStopsValidator that has been set up with a project only
     public SharedStopsValidator buildSharedStopsValidator(Feed feed, SQLErrorStorage errorStorage) {
         if (this.project == null) {
```
br648, line 37: All of the `this.` references can be removed from this method.

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
```diff
@@ -24,24 +23,30 @@ public class SharedStopsValidator extends FeedValidator {
...
             throw new RuntimeException("Shared stops validator can not be called because no project has been set!");
         }
-        return new SharedStopsValidator(feed, errorStorage, this.project);
+        if (this.feedId == null) {
+            throw new RuntimeException("Shared stops validator can not be called because no feed ID has been set!");
+        }
+        return new SharedStopsValidator(feed, errorStorage, this.project, this.feedId);
     }
-    public SharedStopsValidator(Feed feed, SQLErrorStorage errorStorage, Project project) {
+    public SharedStopsValidator(Feed feed, SQLErrorStorage errorStorage, Project project, String feedId) {
```
br648, line 45: The other constructor already defines the `project` and `feedId`. I don't think it is needed here.

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
```diff
@@ -372,7 +373,14 @@ public void validate(MonitorableJob.Status status) {
...
+                Object gtfsFeedId = feedFetcher.getResults(this.namespace, null, null).get(0).get("feed_id");
```
br648, line 387: I take it the feed_info will only ever have one row?

br648 (changes requested):
I don't really see an alternative to using JDBCFetcher (this is the first use of it in DT). I guess the validation cannot be done in GTFS lib because you need the project's shared stop config and this transcends all feed versions so it can't be included there?

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
miles-grant-ibigroup: The distinction is intentional. It is set to null when there is no value found and "" if we are confident there is no value. This might be me being a js developer shining through haha. looking for the null/undefined distinction

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
miles-grant-ibigroup: I get variable name clash if I do this

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
miles-grant-ibigroup: This is the hack we need to use to do the half-instantiation

miles-grant-ibigroup: Since gtfs-lib uses the other constructor, we need to make sure that we allow instantiation of an already instantiated class

src/main/java/com/conveyal/datatools/manager/models/FeedVersion.java
miles-grant-ibigroup: I think if there are multiple we are ok just taking the first

miles-grant-ibigroup:
> I don't really see an alternative to using JDBCFetcher (this is the first use of it in DT). I guess the validation cannot be done in GTFS lib because you need the project's shared stop config and this transcends all feed versions so it can't be included there?

Yeah we'd do all the validation in gtfs-lib but we don't have access to the feed info table in there

br648:
> > I don't really see an alternative to using JDBCFetcher (this is the first use of it in DT). I guess the validation cannot be done in GTFS lib because you need the project's shared stop config and this transcends all feed versions so it can't be included there?
> 
> Yeah we'd do all the validation in gtfs-lib but we don't have access to the feed info table in there

Feed Info is loaded by GTFS-Lib it's just not made available in Feed. This could be updated to include feed info. But if I understand this correctly, GTFS Lib doesn't have the shard stop config so you'd still have a similar issue?

br648 (approved):
Happy to approve after discussing with Miles. This approach keeps GTFS-Lib clean.

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
```diff
@@ -56,6 +61,7 @@ public void validate() {
         final int STOP_GROUP_ID_INDEX = 0;
         final int STOP_ID_INDEX = 2;
         final int IS_PRIMARY_INDEX = 3;
+        final int FEED_ID_INDEX = 1;
```
binh-dam-ibigroup, line 65: For another PR: could the indices be obtained from the CSV header instead?

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
```diff
@@ -103,8 +110,8 @@ public void validate() {
                 }
 
                 // Check for SS_03 (stop_id referenced doesn't exist)
-                // TODO: CHECK FEED ID (adjust the pre-build constructor to include feed_id)
-                if (!stopIds.contains(stopId)) {
+                // Make sure this error is only returned if we are inside the feed that is being checked
+                if (feedId.equals(sharedStopFeedId) && !stopIds.contains(stopId)) {
                     registerError(NewGTFSError.forFeed(NewGTFSErrorType.SHARED_STOP_GROUP_ENTITY_DOES_NOT_EXIST, stopId));
```
binh-dam-ibigroup, line 116: Can you have this error be reported for the current line (instead of the entire feed) of the shared stops file?

binh-dam-ibigroup (approved):
Works well. A nice-to-have would be to report the line number of the non-existing stop entity found.

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
miles-grant-ibigroup: Good call. Have added a TODO

src/main/java/com/conveyal/datatools/manager/jobs/validation/SharedStopsValidator.java
miles-grant-ibigroup: The shared stops file is not in Postgres so the validator can't talk to it. the `stopId` links it to the stops file

binh-dam-ibigroup: I was thinking of line numbers because the shared stops file is processed line by line using `configReader.readRecord()`.

miles-grant-ibigroup: Yes but unfortunately the validator doesn't know about those

PR #569, opened 2023-10-13
Add auth checks to feed source summary

br648:
### Checklist

### Description

Update to the feed source controller to apply two auth items:
- If a user can manage or view a feed
- If non-admin remove admin labels and notes.

Update to add the organization id and notes to the feed source summary.

The unit tests have been updated to as an admin.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
@@ -441,7 +441,8 @@ public class FeedSourceController {
     protected static List<String> cleanFeedSourceLabelIdsForNonAdmins(List<String> labelIds, boolean isAdmin) {
         // Admin can view all feed labels, but a non-admin should only see those with adminOnly=false.
         return Persistence.labels
-            .getFiltered(PersistenceUtils.applyAdminFilter(in("_id", labelIds), isAdmin)).stream()
+            .getFiltered(PersistenceUtils.applyAdminFilter(in("_id", labelIds), isAdmin))
+            .stream()
             .map(label -> label.id)
             .collect(Collectors.toList());
     }
@@ -456,7 +457,8 @@ public class FeedSourceController {
     protected static List<String> cleanFeedSourceNotesForNonAdmins(List<String> noteIds, boolean isAdmin) {
         // Admin can view all feed notes, but a non-admin should only see those with adminOnly=false.
         return Persistence.notes
-            .getFiltered(PersistenceUtils.applyAdminFilter(in("_id", noteIds), isAdmin)).stream()
+            .getFiltered(PersistenceUtils.applyAdminFilter(in("_id", noteIds), isAdmin))
+            .stream()
             .map(note -> note.id)
             .collect(Collectors.toList());
     }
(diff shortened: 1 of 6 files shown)


philip-cline (approved):
Tested and looks good, thanks for the fix. We don't need the note IDs in the `feedSourceSummary` bc we hit the feedsource endpoint when a feedsource is selected, but  it's not the worst thing to keep them around for the future.

src/main/java/com/conveyal/datatools/manager/models/Label.java
```diff
@@ -39,15 +39,24 @@ public String organizationId () {
...
     public Label (String name, String description, String color, boolean adminOnly, String projectId) {
+        this(null, name, description, color, adminOnly, projectId);
+    }
+
+    /**
+     * Create a new label with provided id.
+     */
+    public Label (String id, String name, String description, String color, boolean adminOnly, String projectId) {
         super();
+        if (id != null) {
```
binh-dam-ibigroup: Nit: I'd think the ctor with more params would call the ctor with fewer params.

src/main/java/com/conveyal/datatools/manager/models/Note.java
```diff
@@ -32,6 +32,21 @@ public class Note extends Model implements Serializable {
     /** Whether the note should be visible to project admins only */
     public boolean adminOnly;
 
+    /**
+     * Create a new note with provided id.
+     */
+    public Note(String id, String body, boolean adminOnly) {
+        super();
+        if (id != null) {
```
binh-dam-ibigroup: Is there a need for a ctor that doesn't take an id?

src/test/java/com/conveyal/datatools/manager/controllers/api/FeedSourceControllerTest.java
```diff
@@ -461,10 +509,33 @@ private static FeedVersion createFeedVersion(String id, String feedSourceId, Loc
...
+     * Helper method to create note.
+     */
+    private static Note createNote(String id, String body) {
+        Note note = new Note(id, body, false);
+        Persistence.notes.create(note);
+        return note;
+    }
+
+    /**
+     * Helper method to create label. If the id is provided save the label now if not deffer to test to save.
```
binh-dam-ibigroup: Typo "defer"

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
```diff
@@ -366,53 +391,79 @@ public static FeedSource checkFeedSourcePermissions(Request req, FeedSource feed
...
+    }
+
+    /**
+     * Removes labels from a feed that a user is not allowed to view. Returns cleaned notes.
+     *
+     * @param labelIds The labels to clean.
+     * @param isAdmin  Is the user an admin? Changes what is returned.
+     * @return Labels the user is allowed to see.
+     */
+    protected static List<String> cleanFeedSourceLabelIdsForNonAdmins(List<String> labelIds, boolean isAdmin) {
```
binh-dam-ibigroup, line 441: Could `cleanFeedSourceLabelIdsForNonAdmins` and `cleanFeedSourceNotesForNonAdmins` be combined?

binh-dam-ibigroup (approved):
It works indeed!

src/main/java/com/conveyal/datatools/manager/controllers/api/FeedSourceController.java
br648: Although they are similar, the persistence type (labels/notes) is hard coded and can't be abstracted.

PR #577, opened 2023-11-20
support new `Url` config fields

miles-grant-ibigroup:
### Checklist

- [ ] The description lists all relevant PRs included in this release _(remove this if not merging to master)_
- [x] e2e tests are all passing _(remove this if not merging to master)_

### Description

Adds 2 new fields to `Deployment` that allow for downloading the router/build config from a URL. If the URL field is present, the old field is _ignored_, and a new file downloaded instead. Downloaded file is parsed only by Java's bulit-in toString method, so safety-wise we should be ok.

#### Looking for feedback
Should mongo be updated when we download the data? Are errors handled well? How should we test this?

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/Deployment.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/Deployment.java
@@ -51,6 +51,7 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
 import static com.conveyal.datatools.manager.DataManager.getConfigPropertyAsText;
+import static com.conveyal.datatools.manager.utils.HttpUtils.downloadFileFromURL;
 import static com.mongodb.client.model.Filters.and;
 import static com.mongodb.client.model.Filters.eq;
 
@@ -418,49 +419,36 @@ public class Deployment extends Model implements Serializable {
         out.close();
     }
 
-    private byte[] downloadFileFromURL(URL url) throws IOException {
-        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-
-        InputStream stream = null;
-        try {
-            byte[] chunk = new byte[4096];
-            int bytesRead;
-            stream = url.openStream();
-
-            while ((bytesRead = stream.read(chunk)) > 0) {
-                outputStream.write(chunk, 0, bytesRead);
+    /** Download config from provided URL. */
+    public String downloadConfig(String configUrl) throws IOException {
+        if (configUrl != null) {
+            try {
+                return new String(downloadFileFromURL(new URL(configUrl)), StandardCharsets.UTF_8);
+            } catch (IOException e) {
+                String message = String.format("Could not download file from %s.", configUrl);
+                LOG.error(message);
+                throw new IOException(message, e);
             }
-
-        } catch (IOException e) {
-            e.printStackTrace();
-            return new byte[0];
-        } finally {
-            outputStream.close();
-            if (stream != null) stream.close();
         }
-
-        return outputStream.toByteArray();
+        return null;
     }
 
     /** Generate build config for deployment as byte array (for writing to file output stream). */
-    public byte[] generateBuildConfig() {
-        Project project = this.parentProject();
-
-        // If there is a custombuildconfigUrl, set the customBuildConfig based on the URL
-        if (this.customBuildConfigUrl != null) {
-            try {
-                this.customBuildConfig = new String(downloadFileFromURL(new URL(customBuildConfigUrl)), StandardCharsets.UTF_8);
-            } catch (IOException e) {
-                LOG.error("Could not download file from {}", customBuildConfigUrl);
-                throw new RuntimeException(customBuildConfigUrl);
-            }
-        }
-
+    public byte[] generateBuildConfig() throws IOException {
+        customBuildConfig = downloadConfig(customBuildConfigUrl);
         return customBuildConfig != null
             ? customBuildConfig.getBytes(StandardCharsets.UTF_8)
-            : project.buildConfig != null
-                ? writeToBytes(project.buildConfig)
-                : null;
+            : getProjectBuildConfig();
+    }
+
+    /**
+     * If a project build config exists, return this as a byte array, or null if not available.
+     */
+    private byte[] getProjectBuildConfig() {
+        Project project = parentProject();
+        return project.buildConfig != null
+            ? writeToBytes(project.buildConfig)
+            : null;
     }
 
     public String generateBuildConfigAsString() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -414,9 +418,44 @@ public void dump (File output, boolean includeManifest, boolean includeOsm, bool
         out.close();
     }
 
+    private byte[] downloadFileFromURL(URL url) throws IOException {
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+
+        InputStream stream = null;
+        try {
+            byte[] chunk = new byte[4096];
```
philip-cline: Is the 4096 chunk size arbitrary?

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -453,6 +492,16 @@ private <O extends Serializable> String writeToString(O object) {
     public byte[] generateRouterConfig() throws IOException {
         Project project = this.parentProject();
 
+        if (this.customRouterConfigUrl != null) {
+            try {
+                // TODO: should Mongo be updated here?
```
philip-cline: Should this comment be added above the line for `this.customBuildConfig` as well?

philip-cline (approved):
Looks fine to me, two small questions

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
miles-grant-ibigroup: yes

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
miles-grant-ibigroup: The question is targeted at you. Is leaving the comment the best bet do you think?

philip-cline: If not here, where will it be updated?

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
philip-cline: yes

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
miles-grant-ibigroup: Well I think in this line instead of updating the local instantiated Deployment we'd update Mongo

philip-cline: Right I see what you're saying now. Updating the mongo might be more in line with what we do currently elsewhere

br648:
@miles-grant-ibigroup I've made a few code changes, quicker than going back and forth through comments. I hope you don't mind.

I've added a test, it might not be what you are looking for, but will at least test that the download from URL works.

I'm not sure on saving to Mongo. How often will the downloaded config change and how often will it be download?! If you save the config to Mongo, you then have to deal with the logic around downloading a newer config. We can discuss this if you want?

miles-grant-ibigroup:
> @miles-grant-ibigroup I've made a few code changes, quicker than going back and forth through comments. I hope you don't mind.
> 
> I've added a test, it might not be what you are looking for, but will at least test that the download from URL works.
> 
> I'm not sure on saving to Mongo. How often will the downloaded config change and how often will it be download?! If you save the config to Mongo, you then have to deal with the logic around downloading a newer config. We can discuss this if you want?

Thanks for your refactor I for the most part much prefer it! However, does this take into account that if no URL is set the buildConfig/routerConfig from Mongo should be used?

br648:
> > @miles-grant-ibigroup I've made a few code changes, quicker than going back and forth through comments. I hope you don't mind.
> > I've added a test, it might not be what you are looking for, but will at least test that the download from URL works.
> > I'm not sure on saving to Mongo. How often will the downloaded config change and how often will it be download?! If you save the config to Mongo, you then have to deal with the logic around downloading a newer config. We can discuss this if you want?
> 
> Thanks for your refactor I for the most part much prefer it! However, does this take into account that if no URL is set the buildConfig/routerConfig from Mongo should be used?

I've made a small change to only assign the downloaded config if available.

miles-grant-ibigroup:
> > > @miles-grant-ibigroup I've made a few code changes, quicker than going back and forth through comments. I hope you don't mind.
> > > I've added a test, it might not be what you are looking for, but will at least test that the download from URL works.
> > > I'm not sure on saving to Mongo. How often will the downloaded config change and how often will it be download?! If you save the config to Mongo, you then have to deal with the logic around downloading a newer config. We can discuss this if you want?
> > 
> > 
> > Thanks for your refactor I for the most part much prefer it! However, does this take into account that if no URL is set the buildConfig/routerConfig from Mongo should be used?
> 
> I've made a small change to only assign the downloaded config if available.

This looks great thanks so much! Things are working for me but would appreciate if someone else gave it a quick check too to make sure it still works

philip-cline:
I can give it a quick check

br648 (approved):
LGTM.

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -418,49 +419,36 @@ public void dump (File output, boolean includeManifest, boolean includeOsm, bool
...
+    /** Download config from provided URL. */
+    public String downloadConfig(String configUrl) throws IOException {
+        if (configUrl != null) {
             try {
-                this.customBuildConfig = new String(downloadFileFromURL(new URL(customBuildConfigUrl)), StandardCharsets.UTF_8);
+                return new String(downloadFileFromURL(new URL(configUrl)), StandardCharsets.UTF_8);
             } catch (IOException e) {
-                LOG.error("Could not download file from {}", customBuildConfigUrl);
-                throw new RuntimeException(customBuildConfigUrl);
+                String message = String.format("Could not download file from %s.", configUrl);
```
philip-cline: Nit: we could say `Could not download *config* file from %s` here since the method is specific.

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -418,49 +419,36 @@ public void dump (File output, boolean includeManifest, boolean includeOsm, bool
...
-        Project project = this.parentProject();
-
-        // If there is a custombuildconfigUrl, set the customBuildConfig based on the URL
-        if (this.customBuildConfigUrl != null) {
+    /** Download config from provided URL. */
+    public String downloadConfig(String configUrl) throws IOException {
+        if (configUrl != null) {
             try {
-                this.customBuildConfig = new String(downloadFileFromURL(new URL(customBuildConfigUrl)), StandardCharsets.UTF_8);
+                return new String(downloadFileFromURL(new URL(configUrl)), StandardCharsets.UTF_8);
```
philip-cline, line 425: Should we do a simple check to see if the downloaded String is a valid JSON? Both types of configs should be right?

philip-cline (approved):
Did some more testing on this, it seems to be working well. I had two small comments on the new code as well.

PR #592, opened 2024-03-26
Add Append Transformation

miles-grant-ibigroup:
### Checklist

### Description

Adds new transformation type similar to replace transformation that appends to file instead of replaces. Appending is done in temporary dir. Unit test is included

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
@@ -1,0 +1,76 @@
+
+package com.conveyal.datatools.manager.models.transform;
+
+import com.conveyal.datatools.common.status.MonitorableJob;
+import com.conveyal.datatools.manager.models.TableTransformResult;
+import com.conveyal.datatools.manager.models.TransformType;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+
+public class AppendToFileTransformation extends ZipTransformation {
+
+    public static AppendToFileTransformation create(String csvData, String table) {
+        AppendToFileTransformation transformation = new AppendToFileTransformation();
+        transformation.csvData = csvData;
+        transformation.table = table;
+        return transformation;
+    }
+
+    @Override
+    public void validateParameters(MonitorableJob.Status status) {
+        if (csvData == null) {
+            status.fail("CSV data must not be null");
+        }
+    }
+
+    @Override
+    public void transform(FeedTransformZipTarget zipTarget, MonitorableJob.Status status) {
+        String tableName = table + ".txt";
+        Path targetZipPath = Paths.get(zipTarget.gtfsFile.getAbsolutePath());
+        try (
+                FileSystem targetZipFs = FileSystems.newFileSystem(targetZipPath, (ClassLoader) null);
+                InputStream inputStream = new ByteArrayInputStream(csvData.getBytes(StandardCharsets.UTF_8));
+        ) {
+            TransformType type = TransformType.TABLE_MODIFIED;
+
+            Path targetTxtFilePath = getTablePathInZip(tableName, targetZipFs);
+
+            final File tempFile = File.createTempFile(tableName + "-temp", ".txt");
+            Files.copy(targetTxtFilePath, tempFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+
+            // Append CSV data into the target file in the temporary copy of file
+            try (OutputStream os = new FileOutputStream(tempFile, true)) {
+                os.write(inputStream.readAllBytes());
+            } catch (Exception e) {
+                status.fail("Failed to write to target file", e);
+            }
+
+            // Copy modified file into zip
+            Files.copy(tempFile.toPath(), targetTxtFilePath, StandardCopyOption.REPLACE_EXISTING);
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    @Override
+    public void transform(FeedTransformZipTarget zipTarget, MonitorableJob.Status status) {
+        String tableName = table + ".txt";
+        Path targetZipPath = Paths.get(zipTarget.gtfsFile.getAbsolutePath());
+        try (
+                FileSystem targetZipFs = FileSystems.newFileSystem(targetZipPath, (ClassLoader) null);
+                InputStream inputStream = new ByteArrayInputStream(csvData.getBytes(StandardCharsets.UTF_8));
```
br648: Bring indent back by one.

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -282,6 +314,10 @@ private static String generateStopsWithCustomFields() {
             + "\n1234567,customValue3,customValue4";
     }
 
+    private static String generateStopRow() {
+        return "\nnew,new,appended stop,,37.06668,-122.07781,,,0,123,,";
```
br648: Will the first row of data always start with a new line character? Knowing that the data is being appended there should probably be a check for this and if it is missing it should be added. I would like to see more than one row appended, perhaps 3 or 4 just to be sure of the new line counter.

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
```diff
@@ -0,0 +1,76 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+
+public class AppendToFileTransformation extends ZipTransformation {
+
+    public static AppendToFileTransformation create(String csvData, String table) {
+        AppendToFileTransformation transformation = new AppendToFileTransformation();
+        transformation.csvData = csvData;
```
br648, line 25: This might be out of scope. Should there be a check to confirm that each row of new data has the correct number of columns matching the provided table?

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
```diff
@@ -0,0 +1,76 @@
...
+            zipTarget.feedTransformResult.tableTransformResults.add(new TableTransformResult(
+                    tableName,
+                    type,
+                    0,
+                    0,
+                    lineCount + 1,
+                    0
+            ));
+        } catch (Exception e) {
+            status.fail("Unknown error encountered while transforming zip file", e);
```
br648, line 78: Should this error message be more specific? If there is an error appending I might then know to check the data I'm attempting to append?

src/test/java/com/conveyal/datatools/manager/jobs/ArbitraryTransformJobTest.java
```diff
@@ -11,6 +11,7 @@
 import com.conveyal.datatools.manager.models.Snapshot;
 import com.conveyal.datatools.manager.models.TableTransformResult;
 import com.conveyal.datatools.manager.models.transform.AddCustomFileFromStringTransformation;
+import com.conveyal.datatools.manager.models.transform.AppendToFileTransformation;
```
br648, line 13: Not a result of your PR, but can you remove all the unused imports please?

br648 (commented):
This is not far off. I think it should be able to handle the missing starting new line character, see comment.

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
miles-grant-ibigroup: The problem is we don't know what table we're going to be modifying. If it comes from the database, we can determine it, but if it's coming from fetch we might not know at this point.

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
miles-grant-ibigroup: This only fires if zipfs has a problem so it means there's a problem with the filesystem/datatools rather than any of the user's data!

miles-grant-ibigroup:
Thanks all changes are addressed!

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public void transform(FeedTransformZipTarget zipTarget, MonitorableJob.Status status) {
+        String tableName = table + ".txt";
+        Path targetZipPath = Paths.get(zipTarget.gtfsFile.getAbsolutePath());
+
+        try (
+            FileSystem targetZipFs = FileSystems.newFileSystem(targetZipPath, (ClassLoader) null);
+            InputStream newLineStream = new ByteArrayInputStream("\n".getBytes(StandardCharsets.UTF_8));
```
br648, line 44: Instead of creating an input stream for a single character, could the new line be prefixed to `csvData` on line 25?

br648 (commented):
Thanks for updating, just one item left to consider then I'll approve.

src/main/java/com/conveyal/datatools/manager/models/transform/AppendToFileTransformation.java
miles-grant-ibigroup: Unfortunately that method is only fired during the unit tests!

br648 (approved):
All good following feedback to comments.

JymDyerIBI (approved):
👍  This looks good to me.

PR #603, opened 2024-11-27
MobilityData Validator v6

miles-grant-ibigroup:
Upgrade!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,7 +262,7 @@
         <dependency>
             <groupId>org.mobilitydata.gtfs-validator</groupId>
             <artifactId>gtfs-validator-main</artifactId>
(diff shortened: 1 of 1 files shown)


br648 (approved):
At what point do we pull the validation from GTFS-Lib and rely on this solely?!

miles-grant-ibigroup:
> At what point do we pull the validation from GTFS-Lib and rely on this solely?!

There's still quite a few validators we'd have to get merged into the mobility data one. That will be a big project but probably one that would be appreciated by the community

clinephi:
The community would definitely appreciate it! :)

binh-dam-ibigroup (approved):
Works!

PR #604, opened 2024-12-13
remove project-wide deployment settings

miles-grant-ibigroup:
There were some project wide deployment settings that were causing more harm than good given how datatools is used these days! These settings are removed in this PR

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/models/Deployment.java
+++ b/src/main/java/com/conveyal/datatools/manager/models/Deployment.java
@@ -446,17 +446,12 @@ public class Deployment extends Model implements Serializable {
     }
 
     public String generateBuildConfigAsString() {
-        if (customBuildConfig != null) return customBuildConfig;
-        return null;
-    }
-
-    /** Convenience method to write serializable object (primarily for router/build config objects) to byte array. */
-    private <O extends Serializable> byte[] writeToBytes(O object) {
         try {
-            return otpConfigMapper.writer().writeValueAsBytes(object);
-        } catch (JsonProcessingException e) {
-            LOG.error("Value contains malformed JSON", e);
-            return null;
+            return new String(generateBuildConfig(), StandardCharsets.UTF_8);
+            // TODO: Correctly generate default build config
+        } catch (Exception e) {
+            LOG.error("Failed to generate build config: ", e);
+            return "";
         }
     }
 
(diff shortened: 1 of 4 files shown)


br648 (approved):
All good.

src/main/java/com/conveyal/datatools/manager/models/Deployment.java
```diff
@@ -442,22 +442,12 @@ public byte[] generateBuildConfig() throws IOException {
...
-        Project project = parentProject();
-        return project.buildConfig != null
-            ? writeToBytes(project.buildConfig)
             : null;
     }
 
     public String generateBuildConfigAsString() {
         if (customBuildConfig != null) return customBuildConfig;
-        return writeToString(this.parentProject().buildConfig);
+        return null;
```
binh-dam-ibigroup: Simplify to `return customBuildConfig;`. You might even be able to inline this method.

src/main/java/com/conveyal/datatools/manager/models/OtpBuildConfig.java
```diff
@@ -11,23 +11,4 @@
 public class OtpBuildConfig implements Serializable {
```
binh-dam-ibigroup, line 13: This is now an empty class, can we remove it without messing up the loading of existing datatools projects?

src/main/java/com/conveyal/datatools/manager/models/OtpRouterConfig.java
```diff
@@ -11,31 +11,5 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class OtpRouterConfig implements Serializable {
```
binh-dam-ibigroup, line 14: This is now an empty class, can we remove it without messing up the loading of existing datatools projects?

binh-dam-ibigroup (approved):
Looks good with some further easy cleanup as indicated.

src/main/java/com/conveyal/datatools/manager/models/OtpRouterConfig.java
miles-grant-ibigroup: Removing it caused some problems with db migration, but I can add a TODO to remove these in the future

binh-dam-ibigroup (approved):

PR #608, opened 2025-03-04
Notify subscribers of auto publish failure

br648:
### Checklist

### Description

Notify subsribers if an auto published feed fails.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/AutoPublishJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/AutoPublishJob.java
@@ -54,10 +54,19 @@ public class AutoPublishJob extends MonitorableJobWithResourceLock<FeedSource> {
             }
         }
 
-        // If validation successful, just execute the feed updating process.
         if (!status.error) {
+            // Validation successful, just execute the feed updating process.
             FeedVersionController.publishToExternalResource(latestFeedVersion);
             LOG.info("Auto-published feed source {} to external resource.", feedSource.id);
+        } else {
+            // Notify feed and project subscribed users of failure.
+            String message = String.format(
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/manager/jobs/AutoPublishJob.java
```diff
@@ -54,10 +54,19 @@ protected void innerJobLogic() throws Exception {
...
         }
 
-        // If validation successful, just execute the feed updating process.
         if (!status.error) {
+            // Validation successful, just execute the feed updating process.
             FeedVersionController.publishToExternalResource(latestFeedVersion);
             LOG.info("Auto-published feed source {} to external resource.", feedSource.id);
+        } else {
+            // Notify feed and project subscribed users of failure.
+            String message = String.format(
```
binh-dam-ibigroup, line 63: This will result in two email notifications being sent (for fetched feed versions only):
1. A useful email with the text outlined in this block,
2. A more generic email that is already being sent, and that reads: "While attempting to process a new feed version for [feed name], an unrecoverable error was encountered. More details: unknown error". I think we can keep the two, although in some future, it will make sense to propagate the error content from this message to the other one. Thoughts?

br648: How quickly are MTC expecting an update? I think for now we go with two email notifications.

binh-dam-ibigroup: We can keep it at two emails for now and iterate based on their feedback.

binh-dam-ibigroup (approved):

JymDyerIBI (approved):
👍  Looks good.

PR #610, opened 2025-03-11
Fix GTFS+ downloads on Firefox

### LINKED ISSUE #609: GTFS+ downloads fails on Firefox
GTFS+ editor state is left uninitialized in Firefox when viewing/editing GTFS+ files.
(This issue does not occur in Chromium/WebKit browsers).

Image

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected  => This PR addresses an issue with MTC
- [na] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [na] The description lists any configuration setting(s) that differ from the default settings

### Description

Fix #609, where the implementation of GTFS+ download was * adding * a `Content-encoding: identity` header in addition to the gzip encoding, instead of overwriting the header.

### DIFF
--- a/src/main/java/com/conveyal/datatools/common/utils/SparkUtils.java
+++ b/src/main/java/com/conveyal/datatools/common/utils/SparkUtils.java
@@ -55,7 +55,7 @@ public class SparkUtils {
         raw.setContentType("application/octet-stream");
         raw.setHeader("Content-Disposition", "attachment; filename=" + filename);
         // Override the gzip content encoding applied to standard API responses.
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Was this due to a dependency upgrade? Or has this never worked?

binh-dam-ibigroup:
> Was this due to a dependency upgrade? Or has this never worked?

I think that never worked.

josh-willis-arcadis (approved):
good catch!

PR #620, opened 2025-04-02
remove dependence on OTP REST API

miles-grant-ibigroup:
We previously used the REST API to determine OTP uptime. This is now corrected and we use the https://docs.opentripplanner.org/en/latest/sandbox/ActuatorAPI/ instead. However, this means the actuator api has to be enabled for deployments to succeed.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
+++ b/src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
@@ -136,7 +136,7 @@ public class MonitorServerStatusJob extends MonitorableJob {
             // become available.
             TimeTracker routerCheckTracker = new TimeTracker(20, TimeUnit.MINUTES);
 
-            String routerUrl = String.join("/", ipUrl, deployment.tripPlannerVersion.equals(TripPlannerVersion.OTP_2) ? "otp/actuators/health" : "otp/routers/default");
+            String routerUrl = String.join("/", ipUrl, TripPlannerVersion.getUptimeCheckUrl(deployment.tripPlannerVersion));
             boolean routerIsAvailable = false;
             while (!routerIsAvailable) {
                 // If the request was successful, the graph build is complete!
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
```diff
@@ -134,7 +135,8 @@ public void jobLogic() {
             // load has completed successfully. Wait a maximum of 20 minutes to load the graph and for the router to
             // become available.
             TimeTracker routerCheckTracker = new TimeTracker(20, TimeUnit.MINUTES);
-            String routerUrl = String.join("/", ipUrl, "otp/routers/default");
+
+            String routerUrl = String.join("/", ipUrl, deployment.tripPlannerVersion.equals(TripPlannerVersion.OTP_2) ? "otp/actuators/health" : "otp/routers/default");
```
br648: DT has `enum TripPlannerVersion` in `Deployment` as well as `public enum OtpVersion`. This might be a good time to combine the two? Or at the very least, update `enum TripPlannerVersion` and assoicate the routerUrl with the version so the info is held in one place (similar to `OtpVersion`).

br648 (changes requested):

src/main/java/com/conveyal/datatools/manager/jobs/MonitorServerStatusJob.java
miles-grant-ibigroup: Yeah I was so confused as to why there are 2... I went for the enum one so we don't have to use magic strings. Maybe it's something to do with otp-runner? I'm nervous to change it

danielhep:
How hard would it be to add a check to the otp config to ensure that actuator API is enabled?

miles-grant-ibigroup:
> How hard would it be to add a check to the otp config to ensure that actuator API is enabled?

It's in otp-config.json which in a lot of cases doesn't even get read by DT at all. It's stored as bytes that get sent to otp-runner

br648 (approved):
Nice!

JymDyerIBI (approved):
👍  Looks good to me.

PR #630, opened 2025-07-09
Auto fetch memory leak

br648:
### Checklist

### Description

Update to close connection and input streams used in the process of fetching a feed version. If there are any issues with the connection e.g. bad SSL, the only potential issue would be not closing the connection (which is addressed here). It is more likely that the issue (if it is in this area) is related to the feed being downloaded and input streams, overtime, not being closed. Again, this is addressed here.

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
+++ b/src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
@@ -139,7 +139,9 @@ public class FeedStore {
         // NOTE: depending on the feed store, there may not be a feedSource provided (e.g., gtfsplus)
         File file = new File(path, id);
         LOG.info("Writing file to {}", file.getAbsolutePath());
-        ByteStreams.copy(inputStream, new FileOutputStream(file));
+        try (InputStream stream = inputStream; FileOutputStream outputStream = new FileOutputStream(file)) {
+            ByteStreams.copy(stream, outputStream);
+        }
         if (feedSource != null && !DataManager.useS3) {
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
Code looks good, I will try deploying this before we merge

br648:
@binh-dam-ibigroup is it worth widening the scope to check that all input streams are closed?

binh-dam-ibigroup:
> @binh-dam-ibigroup is it worth widening the scope to check that all input streams are closed?

Probably. You might have to split PRs depending on how many instances you find.

src/main/java/com/conveyal/datatools/manager/persistence/FeedStore.java
```diff
@@ -139,7 +139,9 @@ public File newFeed (String id, InputStream inputStream, FeedSource feedSource)
         // NOTE: depending on the feed store, there may not be a feedSource provided (e.g., gtfsplus)
         File file = new File(path, id);
         LOG.info("Writing file to {}", file.getAbsolutePath());
-        ByteStreams.copy(inputStream, new FileOutputStream(file));
+        try (InputStream stream = inputStream; FileOutputStream outputStream = new FileOutputStream(file)) {
```
binh-dam-ibigroup, line 142: Nit: I'd prefer the stream to be created and closed in the same method, but that would make things complicated. I see that method `newFeed` is often invoked with a stream created from a file, so having this method accept a file path would make things cleaner, but that would result in a different implementation with a download stream...

binh-dam-ibigroup (approved):

PR #631, opened 2025-07-14
Add build log to deployment notification email

miles-grant-ibigroup:
To accomplish this, I've added a new redirect-capable artifact download endpoint

### DIFF
--- a/src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
+++ b/src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
@@ -103,6 +103,7 @@ public class DeploymentController {
         }
         // If a jobId query param is provided, find the matching job summary.
         String jobId = req.queryParams("jobId");
+
         if (jobId != null) {
             for (DeployJob.DeploySummary summary : deployment.deployJobSummaries) {
                 if (summary.jobId.equals(jobId)) {
@@ -137,12 +138,16 @@ public class DeploymentController {
             region = summaryToDownload.ec2Info == null ? null : summaryToDownload.ec2Info.region;
         }
         AmazonS3URI uri = new AmazonS3URI(uriString);
+
+        // If a redirect query param is provided, find the matching job summary.
+        boolean redirect = Boolean.parseBoolean(req.queryParams("redirect"));
+
         // Assume the alternative role if needed to download the deploy artifact.
         return S3Utils.downloadObject(
             S3Utils.getS3Client(role, region),
             uri.getBucket(),
             String.join("/", uri.getKey(), filename),
-            false,
+            redirect,
             req,
             res
         );
@@ -602,6 +607,8 @@ public class DeploymentController {
         options(apiPrefix + "secure/deployments", (q, s) -> "");
         get(apiPrefix + "secure/deployments/:id/download", DeploymentController::downloadDeployment);
         get(apiPrefix + "secure/deployments/:id/artifact", DeploymentController::downloadBuildArtifact);
+        // This path allows the downloaded artifact file to be named anything
+        get(apiPrefix + "secure/deployments/:id/artifact/:ignored", DeploymentController::downloadBuildArtifact);
         get(apiPrefix + "secure/deployments/:id/ec2", DeploymentController::fetchEC2InstanceSummaries, slimJson::write);
         delete(apiPrefix + "secure/deployments/:id/ec2", DeploymentController::terminateEC2InstanceForDeployment, slimJson::write);
         get(apiPrefix + "secure/deployments/:id", DeploymentController::getDeployment, fullJson::write);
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
```diff
@@ -602,6 +607,8 @@ public static void register (String apiPrefix) {
         options(apiPrefix + "secure/deployments", (q, s) -> "");
         get(apiPrefix + "secure/deployments/:id/download", DeploymentController::downloadDeployment);
         get(apiPrefix + "secure/deployments/:id/artifact", DeploymentController::downloadBuildArtifact);
+        // This path allows the downloaded artifact file to be named anything
+        get(apiPrefix + "secure/deployments/:id/artifact/:ignored", DeploymentController::downloadBuildArtifact);
```
br648: @miles-grant-ibigroup This is confusing me. If the value is ignored, why can't the end point above be used?

miles-grant-ibigroup: Without it, we can't make a call to `.../artifact/blah.json` which means we have to call it as `.../artifact` which means the downloaded file is downloaded as `artifact` which is no good.

br648: ok, that makes sense. Can you update the comment with this explanation please and perhaps call it `:expectedFileName` instead of `:ignored`?

br648 (approved):
Just the one item to address, but approving.

src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
miles-grant-ibigroup: Good idea!

src/main/java/com/conveyal/datatools/manager/controllers/api/DeploymentController.java
```diff
@@ -602,6 +607,9 @@ public static void register (String apiPrefix) {
         options(apiPrefix + "secure/deployments", (q, s) -> "");
         get(apiPrefix + "secure/deployments/:id/download", DeploymentController::downloadDeployment);
         get(apiPrefix + "secure/deployments/:id/artifact", DeploymentController::downloadBuildArtifact);
+        // This path allows the downloaded artifact file to be named anything. Without this line, the downloaded file
+        // will be named `artifact`.
```
binh-dam-ibigroup, line 611: Add to the comment that this endpoint accepts query parameters too.

binh-dam-ibigroup: Do we still need the existing endpoint at `secure/deployments/:id/artifact`?

src/main/java/com/conveyal/datatools/manager/jobs/NotifyUsersForSubscriptionJob.java
```diff
@@ -64,6 +64,13 @@ public void notifyUsers() {
                     deployment.projectId,
                     deployment.id
                 );
+                // Add build log text.
+                html += String.format(
+                        "<p>View <a href='%s/api/manager/secure/deployments/%s/artifact/otp-build.log?filename=otp-build.log&jobId=%s&redirect=true'>deployment otp-build.log</a>.</p>",
```
binh-dam-ibigroup, line 69: Is there a way to deduplicate `otp-build.log` in the URL above?

binh-dam-ibigroup: This link from the notification email is not going to work as is. When entering the URL in the browser (same as clicking the link), the response is
```
{"result":"ERR","message":"Authorization header is missing.","code":401,"detail":null}
```
Instead, we probably the link in the email to point to a login-protected page in datatools UI that triggers the download with the required headers for the endpoint.

binh-dam-ibigroup (changes requested):
The link alone is not going to work as written. I think it needs to go to datatools UI first where the download can be triggered.

src/main/java/com/conveyal/datatools/manager/jobs/NotifyUsersForSubscriptionJob.java
miles-grant-ibigroup: If you're already logged into datatools, the link will work. We should think of the use cases

binh-dam-ibigroup: This link alone, even if logged into datatools, does not provide the required headers.

miles-grant-ibigroup: This does not align with what I've seen. I've added a note for us all to discuss this further

PR #656, opened 2026-07-07
tweak java args to run out of memory more elegantly

miles-grant-ibigroup:
Adds new java flags to better handle OOM issues. Lowers the cap to make more room for the heap, and ensures that the app will crash instead of freezing. Finally, makes sure we get a dump we can analyze when it does crash


br648 (approved):
Looks good. I assume there is enough space for the heap dump?

miles-grant-ibigroup:
Yeah the disk should have lots of room

daniel-heppner-ibigroup (approved):
beautiful

